PR #6, opened 2019-02-26
Bugfix- Put Un/Safe Migrations in inheritence chain for all compatibi…

celeen:
…lity modules

We uncovered a bug in `pg_ha_migrations` where Rails Compatibility modules/classes were being skipped for a couple of different reasons.

This PR:
- adds Safe/Unsafe Migrations modules to the inheritance chain inbetween Rails Compatibility Classes. We tried adding these modules once, but ran into test ordering problems when example classes are inherited from different Migrations versions-- specifically, when the first migration version to be included in the test was `V5_1`, a subsequent test using a migration inheriting from `V4_2` did not have access to the `pg_ha_migrations` modules *before* the usual rails migrations methods.
- Redirects unsafe migrations methods to `self.superclass` rather than directly to the Pg connection.  (an underlying assumption here is that given the typical usage of this gem, `self` in that context will be a migration file that is inheriting from some version of `ActiveRecord::Migration`)
- a one-off spec fix that was somehow missed in the last commit; in rails 5.2, the class `IndexDefinition` was changed [from a constant containing a struct](https://github.com/rails/rails/blob/v5.1.6.1/activerecord/lib/active_record/connection_adapters/abstract/schema_definitions.rb) to [a proper class](https://github.com/rails/rails/blob/v5.2.0/activerecord/lib/active_record/connection_adapters/abstract/schema_definitions.rb), meaning it no longer has a `to_h` method.

### DIFF
--- a/lib/pg_ha_migrations/unsafe_statements.rb
+++ b/lib/pg_ha_migrations/unsafe_statements.rb
@@ -4,7 +4,7 @@ module PgHaMigrations::UnsafeStatements
       arg_list = args.map { |arg| arg.inspect }.join(', ')
       # say_with_time args taken from https://github.com/rails/rails/blob/4-2-stable/activerecord/lib/active_record/migration.rb#L654
       say_with_time "#{method_name}(#{arg_list})" do
-        connection.send(method_name, *args, &block)
+        self.class.superclass.send(method_name, *args, &block)
       end
     end
   end
@@ -75,5 +75,3 @@ module PgHaMigrations::UnsafeStatements
     raise PgHaMigrations::UnsafeMigrationError.new(":add_foreign_key is NOT SAFE! Explicitly call :unsafe_add_foreign_key only if you have guidance from a migration reviewer in #service-app-db.")
   end
 end
-
-ActiveRecord::Migration.send(:prepend, PgHaMigrations::UnsafeStatements)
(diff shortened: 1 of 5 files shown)


lib/pg_ha_migrations/unsafe_statements.rb
```diff
@@ -4,7 +4,7 @@ def self.delegate_unsafe_method_to_connection(method_name)
       arg_list = args.map { |arg| arg.inspect }.join(', ')
       # say_with_time args taken from https://github.com/rails/rails/blob/4-2-stable/activerecord/lib/active_record/migration.rb#L654
       say_with_time "#{method_name}(#{arg_list})" do
```
jcoleman: We might be able to avoid the need for `say_with_time` if we are calling the migration class instead of the connection?

spec/safe_migration_spec.rb
```diff
@@ -527,7 +527,7 @@ def up
 
             indexes = ActiveRecord::Base.connection.indexes("foos")
             expect(indexes.size).to eq(1)
-            expect(indexes.first.to_h).to include(:table => "foos", :name => "index_foos_on_bar", :columns => ["bar"])
```
jcoleman: Could we break this out as a separate commit?

jcoleman: Heh, didn't see this was already merged :D

celeen: :/ sorry, I probably should have waited

PR #7, opened 2019-02-27
Disable DDL transactions by default

celeen:
@JonMR and I were working a few weeks ago on adding this gem to an existing application. We came across the realization that this library should disable DDL transactions by default.

ActiveRecord uses DDL transactions by default; this means that every migration is wrapped in a transaction. If the migration in question alters multiple objects, it will likely acquire exclusive locks on all of those objects, all within one transaction, which can decrease availability in non-trivial ways. Rather than attempt to track or prevent a single migration with "too many" objects altered, we opt to avoid running migrations in a transaction by setting ActiveRecord's `disable_ddl_transations` to true by default. [This blog post has more detailed information](https://medium.com/braintree-product-technology/postgresql-at-scale-database-schema-changes-without-downtime-20d3749ed680).

This change also makes running `safe_add_concurrent_index` possible without having to manually disable ddl transactions on a per migration basis.

One this to bear in mind, however, is that with ddl transactions disabled by default, the changeset in one migration is no longer atomic. This allows a migration to fail part-way through.

An application can override our library's default, setting it back to *enabling* ddl transactions; we have added a reference to how to do so in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,8 @@ Migrations prefixed with `unsafe_` will warn when invoked. The API is designed t
 
 Migrations prefixed with `safe_` prefer concurrent operations where available, set low lock timeouts where appropriate, and decompose operations into multiple safe steps.
 
+[Running multiple DDL statements inside a transaction acquires exclusive locks on all of the modified objects](https://medium.com/braintree-product-technology/postgresql-at-scale-database-schema-changes-without-downtime-20d3749ed680#cc22). For that reason, this gem [disables DDL transactions](./lib/pg_ha_migrations.rb:8) by default. You can change this by resetting `ActiveRecord::Migration.disable_ddl_transaction` in your application.
+
 The following functionality is currently unsupported:
 
 - Rollbacks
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -52,6 +52,8 @@ Migrations prefixed with `unsafe_` will warn when invoked. The API is designed t
 
 Migrations prefixed with `safe_` prefer concurrent operations where available, set low lock timeouts where appropriate, and decompose operations into multiple safe steps.
 
+Default behavior of this gem [disables ddl transactions](./lib/pg_ha_migrations.rb:8). You can change this by resetting `ActiveRecord::Migration.disable_ddl_transactions` in your application.
```
jcoleman: Maybe add as a prefix to this paragraph:

> Because [running multiple DDL statements inside a transaction...acquire[s] exclusive locks on all of [the modified] objects](https://medium.com/braintree-product-technology/postgresql-at-scale-database-schema-changes-without-downtime-20d3749ed680#cc22), we default to [disabling DDL transactions...]

lib/pg_ha_migrations/hacks/disable_ddl_transaction.rb
```diff
@@ -0,0 +1,13 @@
+require "active_record/migration"
+
+module ActiveRecordHack
```
jcoleman: Could we put this inside `PgHaMigrations` instead?

README.md
jcoleman: `disable_ddl_transactions` should be singular, I believe.

Issue #17, opened 2019-03-07
Make monkey-patching of ActiveRecord::Migration methods optional


tragiclifestories:
Really glad this exists - thanks people!

One thing that would make things easier is if it was possible not to replace the original implementation of the various ActiveRecord::Migration DDL methods with versions that raise errors. The use case for this is incrementally rolling this out in an organisation with a big old monolith with multiple teams etc. rolling it out. We'd like to have the choice ideally.

Happy to do a PR if one would be accepted.

redneckbeard:
Hey James -- we'd be okay with that if it were a configurable option and
not the default.

PR #18, opened 2019-03-14
Check for dependent indexes before dropping columns

jcoleman:
We generally have two major types of concerns we care about for the
safe/unsafe distinction:
- Database safety (e.g., long-held locks)
- Application safety (e.g., dropping columns the app uses)

Currently we check for both of those concerns in the `safe_*` method
variants, and the `unsafe_*` variants are always a direct dispatch to
the native ActiveRecord migration method.

But it seems there's a class of problems that we can't handle easily
without breaking that design rule a bit. For example, dropping a column
is unsafe from an application perspective, so we want to call it out
explicitly with an `unsafe_` prefix to give it extra visibility (someone
needs to go confirm the application is rollback safe without this
column present) -- that means there is no `safe_remove_column`. However
there are other conditions we'd like to assert before dropping a column:
for example, dropping an unused column that's contained in one or more
indexes may be safe from an application perspective, but the cascading
drop of the index won't use the `CONCURRENT` operation form and is
therefore unsafe from a database perspective.

In this changeset I've relaxed the "`unsafe_*` variants always directly
dispatch to the native AR method" rule a bit and introduced the concept
of "dependent object checks". For this first iteration I've decided to
only handle the `remove_column` case where at least one index depends on
the column being removed, but I've tried to ensure it's reasonably
extensible.

Still TODO: I need to add a mechanism for making this behavior opt-in
(at least initially) for the 1.x series since it's not backwards
compatible with some existing migrations. Additionally I need to add
 documentation for the feature and add a test for arg parsing.


lib/pg_ha_migrations/dependent_objects_checks.rb
```diff
@@ -0,0 +1,60 @@
...
+            'index' AS dependent_type,
+            dep_class.relname AS dependent_name
+          FROM pg_catalog.pg_depend
+          JOIN pg_catalog.pg_class ref_class ON ref_class.oid = pg_depend.refobjid
+          JOIN pg_catalog.pg_attribute ref_attr ON ref_attr.attrelid = ref_class.oid
+            AND ref_attr.attnum = pg_depend.refobjsubid
+          JOIN pg_catalog.pg_class dep_class ON dep_class.oid = pg_depend.objid
+          WHERE pg_depend.deptype = 'a'
+            AND pg_depend.refclassid = 'pg_class'::regclass
+            AND ref_class.relname = '#{table_name}'
```
jcoleman: I should clean up the interpolation here.

jcoleman:
@tragiclifestories Thanks for the `config` infrastructure in #17; I've utilized that here also.

tragiclifestories:
No problem - I have further plans for it myself 😉

This PR is interesting. When we were scoping what would have been basically an equivalent to this gem, we thought it would be good to have a mechanism to check for situations where a usually unsafe migration was safe - when, for example, a table was empty and you were adding a cooling with a

README.md
```diff
@@ -48,11 +48,23 @@ and never use `def change`. We believe that this is the only safe approach in pr
...
+
+The API is designed to be explicit yet remain flexible. There may be situations where invoking the `unsafe_*` method is preferred (or the only option available for definitionally unsafe operations).
+
+While `unsafe_*` methods were historically (through 1.0) pure wrappers for invoking the native ActiveRecord migration method, there is a class of problems that we can't handle easily without breaking that design rule a bit. For example, dropping a column is unsafe from an application perspective, so we make the application safety concerns explicit by using an `unsafe_` prefix. Using `unsafe_remove_column` calls out the need to audit the application to confirm the migration won't break the application. Because there are no safe alternatives we don't define a `safe_remove_column` analogue. However there are still conditions we'd like to assert before dropping a column. For example, dropping an unused column that's used in one or more indexes may be safe from an application perspective, but the cascading drop of the index won't use a `CONCURRENT` operation drop the dependent indexes and is therefore unsafe from a database perspective.
```
tragiclifestories:
```suggestion
While `unsafe_*` methods were historically (through 1.0) pure wrappers for invoking the native ActiveRecord migration method, there is a class of problems that we can't handle easily without breaking that design rule a bit. For example, dropping a column is unsafe from an application perspective, so we make the application safety concerns explicit by using an `unsafe_` prefix. Using `unsafe_remove_column` calls out the need to audit the application to confirm the migration won't break the application. Because there are no safe alternatives we don't define a `safe_remove_column` analogue. However there are still conditions we'd like to assert before dropping a column. For example, dropping an unused column that's used in one or more indexes may be safe from an application perspective, but the cascading drop of the index won't use a `CONCURRENT` operation to drop the dependent indexes and is therefore unsafe from a database perspective.
```

tragiclifestories: This is the world's most useless diff - the change is "won't use a `CONCURRENT` operation drop the dependent indexes" -> "won't use a `CONCURRENT` operation drop *to* the dependent indexes"

jcoleman:
> check for situations where a usually unsafe migration was safe - when, for example, a table was empty and you were adding a column with a non null default. We thought you could add a sort of preflight check to see that the table was empty or near enough.

@tragiclifestories Yep, we've talked about this before a little bit; our current approach is to use the `unsafe_` variant, include a comment with row counts, and have someone sign off on it being safe. This retains the explicitness of denoting a generally unsafe change but allows us to proceed with caution as needed.

One of the difficulties to automating that kind of check is that it's entirely possible the first time an error would occur is in your production deploy (since row counts are almost certainly low in development).

README.md
```diff
@@ -48,11 +48,23 @@ and never use `def change`. We believe that this is the only safe approach in pr
...
+
+When `unsafe_*` migration methods support checks of this type you can bypass the checks by passing an `:allow_dependent_objects` key in the method's `options` hash containing an array of dependent object types you'd like to allow. Until 2.0 none of these checks will run by default, but you can opt-in by setting `config.check_for_dependent_objects = true` [in your configuration initializer](#configuration).
```
JonMR: Just verifying my understanding. Once 2.0 is out, users will need to either pass the `:allow_dependent_objects` option or set `config.check_for_dependent_objcects = false`?

jcoleman: Yep, in 2.0+ I'd plan to make these checks the default behavior (but still be able to disable that).

PR #19, opened 2019-03-15
Allow users to re-enable default migration methods

### LINKED ISSUE #17: Make monkey-patching of ActiveRecord::Migration methods optional

tragiclifestories:
Fixes #17.

It turns out that looking up the default implementation for this purpose is quite as tricky as it is when aliasing to `unsafe_$FOO` so the implementation is essentially to introduce a very similar method to `delegate_unsafe_method_to_migration_base_class` and pull out the shared part to a helper.

There's a couple of failing tests but these seem to be also failing on master (related to creating indexes concurrently) and unrelated to this PR.

### DIFF
--- a/lib/pg_ha_migrations.rb
+++ b/lib/pg_ha_migrations.rb
@@ -5,6 +5,16 @@ require "active_record/migration"
 require "relation_to_struct"
 
 module PgHaMigrations
+  Config = Struct.new(:disable_default_migration_methods)
+
+  def self.config
+    @config ||= Config.new(true)
+  end
+
+  def self.configure
+    yield config
+  end
+
   LOCK_TIMEOUT_SECONDS = 5
   LOCK_FAILURE_RETRY_DELAY_MULTLIPLIER = 5
 
--- a/lib/pg_ha_migrations/unsafe_statements.rb
+++ b/lib/pg_ha_migrations/unsafe_statements.rb
@@ -1,114 +1,93 @@
 module PgHaMigrations::UnsafeStatements
+  def self.disable_or_delegate_default_method(method_name, error_message, allow_reentry_from_compatibility_module: false)
+    define_method(method_name) do |*args, &block|
+      if PgHaMigrations.config.disable_default_migration_methods
+        # Most migration methods are only ever called by a migration and
+        # therefore aren't re-entrant or callable from another migration
+        # method, but `execute` is called directly by at least one of the
+        # implementations in `ActiveRecord::Migration::Compatibility` so
+        # we have to explicitly handle that case by allowing execution of
+        # the original implementation by its original name.
+        unless  allow_reentry_from_compatibility_module && caller[0] =~ /lib\/active_record\/migration\/compatibility.rb/
+          raise PgHaMigrations::UnsafeMigrationError, error_message
+        end
+      end
+
+      execute_ancestor_statement(method_name, *args, &block)
+    end
+  end
+
   def self.delegate_unsafe_method_to_migration_base_class(method_name)
     define_method("unsafe_#{method_name}") do |*args, &block|
-      # Dispatching here is a bit complicated: we need to execute the method
-      # belonging to the first member of the inheritance chain (besides
-      # UnsafeStatements). If don't find the method in the inheritance chain,
-      # we need to rely on the ActiveRecord::Migration#method_missing
-      # implementation since much of ActiveRecord::Migration's functionality
-      # is not implemented in real methods but rather by proxying.
-      #
-      # For example, ActiveRecord::Migration doesn't define #create_table.
-      # Instead ActiveRecord::Migration#method_missing proxies the method
-      # to the connection. However some migration compatibility version
-      # subclasses _do_ explicitly define #create_table, so we can't rely
-      # on only one way of finding the proper dispatch target.
-
-      # Exclude our `raise` guard implementations.
-      ancestors_without_unsafe_statements = self.class.ancestors - [PgHaMigrations::UnsafeStatements]
-
-      delegate_method = self.method(method_name)
-      candidate_method = delegate_method
-
-      # Find the first usable method in the ancestor chain
-      # or stop looking if there are no more possible
-      # implementations.
-      until candidate_method.nil? || ancestors_without_unsafe_statements.include?(candidate_method.owner)
-        candidate_method = candidate_method.super_method
-      end
-
-      if candidate_method
-        delegate_method = candidate_method
-      end
-
-      # If we failed to find a concrete implementation from the
-      # inheritance chain, use ActiveRecord::Migrations# method_missing
-      # otherwise use the method from the inheritance chain.
-      if delegate_method.owner == PgHaMigrations::UnsafeStatements
-        method_missing(method_name, *args, &block)
-      else
-        delegate_method.call(*args, &block)
-      end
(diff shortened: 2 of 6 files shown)


spec/spec_helper.rb
```diff
@@ -16,19 +16,19 @@
 
   config.before(:all) do
     ActiveRecord::Base.connection.tables.each do |table|
-      ActiveRecord::Base.connection.execute("DROP TABLE #{table}")
+      ActiveRecord::Base.connection.execute("DROP TABLE #{table} CASCADE")
```
tragiclifestories: These changes are necessary to properly clean up the database after the new tests I've introduced (the foreign key related ones create dependent objects that block dropping the tables).

lib/pg_ha_migrations/unsafe_statements.rb
```diff
@@ -1,114 +1,85 @@
...
-      end
+      execute_ancestor_statement(method_name, *args, &block)
     end
   end
 
-  delegate_unsafe_method_to_migration_base_class :create_table
-  def create_table(name, options={})
-    raise PgHaMigrations::UnsafeMigrationError.new(":create_table is NOT SAFE! Use safe_create_table instead")
+  def self.delegate_unsafe_method_to_migration_base_class(method_name)
+    define_method("unsafe_#{method_name}") { |*args, &block| execute_ancestor_statement(method_name, *args, &block) }
```
jcoleman:
```suggestion
    define_method("unsafe_#{method_name}") do |*args, &block|
      execute_ancestor_statement(method_name, *args, &block)
    end
```

lib/pg_ha_migrations/unsafe_statements.rb
```diff
@@ -1,114 +1,85 @@
...
-      until candidate_method.nil? || ancestors_without_unsafe_statements.include?(candidate_method.owner)
-        candidate_method = candidate_method.super_method
-      end
-
-      if candidate_method
-        delegate_method = candidate_method
+  def self.disable_or_delegate_default_method(method_name, error_message, delegate_from_compatibility: false)
+    define_method(method_name) do |*args, &block|
+      if PgHaMigrations.config.disable_default_migration_methods
+        unless  delegate_from_compatibility && caller[0] =~ /lib\/active_record\/migration\/compatibility.rb/
```
jcoleman: Perhaps instead of `delegate_from_compatibility` this would be better named something like `allow_reentry_from_compatibility_module`?

Also, I realize this was originally my code (and I should have done this earlier), but could we add a comment like:

```
# Most migration methods are only ever called by a migration and
# therefore aren't re-entrant or callable from another migration
# method, but `execute` is called directly by at least one of the
# implementations in `ActiveRecord::Migration::Compatibility` so
# we have to explicitly handle that case by allowing execution of
# the original implementation by its original name.
```

spec/safe_migration_spec.rb
```diff
@@ -43,165 +43,336 @@ def up
...
         end
 
-        it "raises when using default change_column method" do
-          migration = Class.new(migration_klass) do
-            def up
-              safe_create_table :foos
-              safe_add_column :foos, :bar, :text
-              change_column :foos, :bar, :string
+        context "when not configured to disable default migration methods" do
+          before do
```
jcoleman:
```suggestion
          before(:each) do
```

spec/safe_migration_spec.rb
```diff
@@ -43,165 +43,336 @@ def up
...
+            allow(PgHaMigrations.config).
```
jcoleman: Our house style is to using leading rather than trailing dots; could you update?

lib/pg_ha_migrations.rb
```diff
@@ -5,6 +5,16 @@
 require "relation_to_struct"
 
 module PgHaMigrations
+  Config = Struct.new(:disable_default_migration_methods)
```
jcoleman: I'd like to see a couple of tests explicitly of the configuration setting and defaults.

tragiclifestories:
@jcoleman Thanks for the prompt review. I think that's everything - config tests are in `pg_ha_migrations_spec.rb`.

spec/pg_ha_migrations_spec.rb
```diff
@@ -5,6 +5,30 @@
     expect(ActiveRecord::Migration.disable_ddl_transaction).to be_truthy
   end
 
+  describe "config" do
+    before(:each) do
+      PgHaMigrations.instance_variable_set(:@config, nil)
```
jcoleman: Because this runs only as a `before`, it's possible for lines 21-25 to pollute future tests.

The general RSpec pattern for this is to duplicate this into both a `before(:all)` (which handles the first test example that runs) and a `after(:each)` (which ensures every subsequent test is safe). Given that this is just an ivar that will default to not being set anyway, I think just `after(:each)` would be sufficient.

spec/pg_ha_migrations_spec.rb
```diff
@@ -5,6 +5,30 @@
...
+    end
+
+    context "disable_default_migration_methods" do
+      subject(:value) { PgHaMigrations.config.disable_default_migration_methods }
```
jcoleman: We'd prefer to have these two tests assert directly on the config method instead of pulling it out into a `subject` (this is a bit of house style request again 😄).

spec/pg_ha_migrations_spec.rb
```diff
@@ -5,6 +5,36 @@
     expect(ActiveRecord::Migration.disable_ddl_transaction).to be_truthy
   end
 
+  describe "config" do
+    after(:each) do
+      PgHaMigrations.instance_variable_set(:@config, nil)
+    end
+
+    context "disable_default_migration_methods" do
+      subject(:value) { }
```
jcoleman: This is extraneous now?

spec/pg_ha_migrations_spec.rb
```diff
@@ -5,6 +5,36 @@
...
+      context "overridden to be false" do
+        before(:each) do
+          PgHaMigrations.configure do |config|
+            config.disable_default_migration_methods = false
+          end
+        end
+
+        it "is set to false" do
+          expect(PgHaMigrations.config.disable_default_migration_methods)
+            .to be false
```
jcoleman:
```suggestion
            .to be(false)
```

spec/pg_ha_migrations_spec.rb
```diff
@@ -5,6 +5,36 @@
...
+      PgHaMigrations.instance_variable_set(:@config, nil)
+    end
+
+    context "disable_default_migration_methods" do
+      subject(:value) { }
+
+      context "by default" do
+        it "is set to true" do
+          expect(PgHaMigrations.config.disable_default_migration_methods)
+            .to be true
```
jcoleman:
```suggestion
            .to be(true)
```

spec/pg_ha_migrations_spec.rb
tragiclifestories: 🤦‍♂️

PR #21, opened 2019-03-25
Add configuration option for disabling DDL transactions

tragiclifestories:
The gem, as currently implemented, disables all DDL transactions by default, for reducing the possibility of lock problems. We actually prefer to keep the DDL transactions to avoid situations where migrations are only half-applied, and instead encourage more smaller migrations through code review. This PR makes this behaviour opt-out. 

I've also refactored the `Config` into a proper class so it accepts keyword arguments, to avoid the situation we're getting into with `Config.new(true, false, true ... )`

### DIFF
--- a/lib/pg_ha_migrations.rb
+++ b/lib/pg_ha_migrations.rb
@@ -7,6 +7,7 @@ require "relation_to_struct"
 
 module PgHaMigrations
   DEFAULT_CONFIG = {
+    disable_ddl_transactions: true,
     disable_default_migration_methods: true,
   }
 
--- a/lib/pg_ha_migrations.rb
+++ b/lib/pg_ha_migrations.rb
@@ -1,14 +1,17 @@
 require "pg_ha_migrations/version"
+require "pg_ha_migrations/config"
 require "rails"
 require "active_record"
 require "active_record/migration"
 require "relation_to_struct"
 
 module PgHaMigrations
-  Config = Struct.new(:disable_default_migration_methods)
+  DEFAULT_CONFIG = {
+    disable_default_migration_methods: true,
+  }
 
   def self.config
-    @config ||= Config.new(true)
+    @config ||= Config.new(**DEFAULT_CONFIG)
   end
 
   def self.configure
--- a/lib/pg_ha_migrations/config.rb
+++ b/lib/pg_ha_migrations/config.rb
@@ -1,0 +1,9 @@
+module PgHaMigrations
+  class Config
+    attr_accessor :disable_default_migration_methods
+
+    def initialize(disable_default_migration_methods:)
+      @disable_default_migration_methods = disable_default_migration_methods
+    end
+  end
+end
--- a/lib/pg_ha_migrations/config.rb
+++ b/lib/pg_ha_migrations/config.rb
@@ -1,9 +1,12 @@
 module PgHaMigrations
   class Config
-    attr_accessor :disable_default_migration_methods
+    attr_accessor :disable_default_migration_methods,
+                  :disable_ddl_transactions
 
-    def initialize(disable_default_migration_methods:)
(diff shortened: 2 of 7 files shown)


lib/pg_ha_migrations.rb
```diff
@@ -1,14 +1,18 @@
...
 module PgHaMigrations
-  Config = Struct.new(:disable_default_migration_methods)
+  DEFAULT_CONFIG = {
+    disable_ddl_transactions: true,
+    disable_default_migration_methods: true,
+  }
 
   def self.config
-    @config ||= Config.new(true)
+    @config ||= Config.new(**DEFAULT_CONFIG)
```
JonMR: Nice, I like this change.

lib/pg_ha_migrations/config.rb
```diff
@@ -0,0 +1,12 @@
+module PgHaMigrations
+  class Config
```
JonMR: 👍 This will be much more readable.

spec/safe_migration_spec.rb
```diff
@@ -1278,5 +1297,52 @@ def up
...
       end
     end
+
+    describe "disable_ddl_transaction" do
+      context "when instance variable is unset" do
+        let(:migration) do
+          Class.new(migration_klass).new
+        end
+
+        context "when default is true" do
```
JonMR: Small thing here, but I was thinking the word `config` or or `configuration` would be clearer than `default`. Since the term `default` is already used in the README and config to denote that it's set to `false` out of the box.

tragiclifestories: I've gone for `"when configured globally to be (true|false)"`

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -104,6 +104,12 @@ def exec_migration(conn, direction)
 
   def safely_acquire_lock_for_table(table, &block)
     _check_postgres_adapter!
+
+    # This method is NOT safe to run more than once in a DDL transaction,
+    # as it could potentially lead to multiple tables being locked
+    # indefinitely
+    return block.call unless disable_ddl_transaction
```
jcoleman: I'm not sure this is correct; if we allow DDL transactionality then the locking problem exists whether or not we use explicit locking with this method or the locks automatically acquired by any DDL statement in Postgres.

Because it's not possible to avoid locks altogether, allowing DDL transactionality across multiple statements (single statements are already inherently transactional) necessarily means that you have to accept the risk of a serious outage (because of the possibility of deadlocks, etc.)

The one exception to this is when multiple changes need to be made against the same object, but in that case I think the appropriate action is to wrap that set of changes in an explicit `transaction do; ...; end` block.

tragiclifestories:
@jcoleman Thanks for looking at this. After looking at the problem and our goals in relation to this again, we've decided that it's probably better that we build an in-house solution. Our own opinions about how to proceed safely clearly diverge from Braintree's in significant ways which would probably be better served by our own implementation rather than trying to foist our use case on pg_ha_migrations.

So far as this one goes, if you want to accept it then I'll respond to the problems raised, fix it up and remove the conflicts, but seeing as only we wanted it, I'm also happy to close it.

jcoleman:
@JonMR and I were looking through open PRs again, so just wanted to loop back around on this: while we would discourage its use, we'd be happy to commit an option to re-enable transactions around whole migrations. But even if we did that it wouldn't be correct to change `safely_acquire_lock_for_table` this way. Postgres is going to acquire locks of the objects being modified with or without it, and those locks won't be released until the transaction is committed. So the "multiple objects being locked" (as well as potential deadlocks) is a problem either way.

JonMR:
I'm going to go ahead and close this for now. If anyone else finds this and has interest in the functionality, we'd be happy to re-open it.

PR #23, opened 2019-03-28
Use locking mechanism as intended in unsafe_add_column

### LINKED ISSUE #22: Definition of `unsafe_add_column` is unused
While trying to test some lock scenarios, I spotted this little problem.

Basically, unlike all the other `unsafe_` methods, this one - along with `unsafe_make_column_not_nullable` - is supposed to have a custom implementation that uses the lock acquisition logic rather than delegating directly to `ActiveRecord::Migration`. However, the custom implementation are further away in the inheritance chain, so it never gets called - the delegate in `unsafe_statements` gets called instead.

I thought initially that this was a bug I'd introduced, but I've reproduced it on 1.0.0 as well. 

Fix should be as simple as removing the delegate from `unsafe_statements` ...  

EDIT: only `unsafe_add_column` is affected, not `unsafe_make_column_not_nullable` as I initially thought.

tragiclifestories:
Fixes #22 .

### DIFF
--- a/lib/pg_ha_migrations/unsafe_statements.rb
+++ b/lib/pg_ha_migrations/unsafe_statements.rb
@@ -24,7 +24,6 @@ module PgHaMigrations::UnsafeStatements
   end
 
   delegate_unsafe_method_to_migration_base_class :create_table
-  delegate_unsafe_method_to_migration_base_class :add_column
   delegate_unsafe_method_to_migration_base_class :change_table
   delegate_unsafe_method_to_migration_base_class :drop_table
   delegate_unsafe_method_to_migration_base_class :rename_table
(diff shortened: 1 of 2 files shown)


lib/pg_ha_migrations/unsafe_statements.rb
```diff
@@ -50,6 +49,12 @@ def self.delegate_unsafe_method_to_migration_base_class(method_name)
   disable_or_delegate_default_method :remove_index, ":remove_index is NOT SAFE! Use safe_remove_concurrent_index instead for Postgres 9.6 databases; Explicitly call :unsafe_remove_index to proceed on Postgres 9.1"
   disable_or_delegate_default_method :add_foreign_key, ":add_foreign_key is NOT SAFE! Explicitly call :unsafe_add_foreign_key only if you have guidance from a migration reviewer in #service-app-db."
 
+  def unsafe_add_column(table, column, type, options = {})
+    safely_acquire_lock_for_table(table) do
```
tragiclifestories: It's a little unfortunate that we now have an implicit dependency in this file on this method in `safe_statements`; but the only alternative I can see is calling `execute_ancestor_statement` in that file, which is not an improvement. Perhaps they could be moved to a separate helper module.

JonMR:
Closing the PR for now since it's been open for so long. If someone else wants to pick this up that'd be great.

This is probably a good starting point for #22.

PR #33, opened 2019-11-04
update rails dependency to include 6.0

roman-melnyk:
update `ALLOWED_VERSIONS` to support `6.0`
update `rails` gem version to  `< 6.1`
update `relation_to_struct` gem version to `>= 1.5.1`
add `rails_6.0` appraisal
fix specs
  - new `timestamps` precision for rails 6 - https://github.com/rails/rails/pull/34970


jcoleman:
Looking at the test run it seems that the test build didn't actually run against Rails 6 (it tested `ActiveRecord::Migration::Compatibility::V5_1` and then `ActiveRecord::Migration::Current` rather than also having `ActiveRecord::Migration::Compatibility::V5_2`).

While we don't want to merge it this way, perhaps to at least get a CI run visible it'd be worth setting the minimum ActiveRecord/Rails version to 6 and see what happens?

roman-melnyk:
I've tested with breakpoint, seems to be tested properly

- [4.2](https://i.imgur.com/f4lBTCT.png)
- [5.0](https://i.imgur.com/MEtEvtV.png)
- [5.1](https://i.imgur.com/qJO79z6.png)
- [5.2](https://i.imgur.com/373C0TZ.png)
- [Current](https://i.imgur.com/vRFT5Rl.png)

JonMR:
:wave: We've been intending to add https://github.com/thoughtbot/appraisal to this project for a while now. I can attempt to take that up over the next few weeks. That would allow us to test against many versions of rails automatically.

I think that's a good step to getting this testable in CI.

gemfiles/rails_6.0.gemfile
```diff
@@ -0,0 +1,7 @@
+# This file was generated by Appraisal
```
stephenreid: 💯

Appraisals
```diff
@@ -10,3 +10,7 @@ appraise "rails-5.2" do
   gem "rails", "5.2.3"
 end
 
+appraise "rails-6.0" do
```
stephenreid: :100

jcoleman:
@roman-melnyk I guess GH didn't email me that you'd updated and rebased this; I'm reviewing now.

lib/pg_ha_migrations/blocking_database_transactions_reporter.rb
```diff
@@ -1,4 +1,5 @@
 require 'stringio'
+require 'active_support/core_ext/string/strip'
```
jcoleman: Instead of adding this line, could you change line 27/28 to use `<<~eos` instead of `<<-eos.strip_heredoc`? This is a language feature since Ruby 2.3, so I think we can simplify.

jcoleman:
@roman-melnyk I've merged; I've just opened another small PR, so will cut a new release either today or tomorrow when that's merged.

jcoleman:
@roman-melnyk I just pushed 1.2.3 to Rubygems.

PR #34, opened 2019-11-21
Allow add_column with default at PG 11

stephenreid:
With PostgreSQL 11, Add Columns including with default values (constant) are now non locking (defaults will be written on page write).
https://www.postgresql.org/docs/11/ddl-alter.html#DDL-ALTER-ADDING-A-COLUMN
https://brandur.org/postgres-default

Test locally in table before and after 15,166 records on small machine

With PG 11
```
=> 15166
➜  ✗ bundle exec rake db:migrate
== 20191125152509 AddColumnToTable: migrating -  ==================
-- add_column(:existing_table, :new_column, :boolean, {:null=>false, :default=>false})
   -> 0.4040s
== 20191125152509 AddColumnToTable: migrated (0.4042s) -  =========
```
With PG 12
```
== 20191125152509 AddColumnToTable: migrating -  ==================
-- add_column(:existing_table, :new_column, :boolean, {:null=>false, :default=>false})
   -> 0.0112s
== 20191125152509 AddColumnToTable migrated (0.0114s) -  =========
```

### DIFF
--- a/lib/pg_ha_migrations/safe_statements.rb
+++ b/lib/pg_ha_migrations/safe_statements.rb
@@ -26,11 +26,13 @@ module PgHaMigrations::SafeStatements
   end
 
   def safe_add_column(table, column, type, options = {})
-    if options.has_key? :default
-      raise PgHaMigrations::UnsafeMigrationError.new(":default is NOT SAFE! Use safe_change_column_default afterwards then backfill the data to prevent locking the table")
-    end
-    if options[:null] == false
-      raise PgHaMigrations::UnsafeMigrationError.new(":null => false is NOT SAFE if the table has data! If you _really_ want to do this, use unsafe_make_column_not_nullable")
+    unless ActiveRecord::Base.connection.postgresql_version >= 110000
+      if options.has_key? :default
+        raise PgHaMigrations::UnsafeMigrationError.new(":default is NOT SAFE! Use safe_change_column_default afterwards then backfill the data to prevent locking the table")
+      end
+      if options[:null] == false
+        raise PgHaMigrations::UnsafeMigrationError.new(":null => false is NOT SAFE if the table has data! If you _really_ want to do this, use unsafe_make_column_not_nullable")
+      end
     end
 
     unsafe_add_column(table, column, type, options)
--- a/lib/pg_ha_migrations/safe_statements.rb
+++ b/lib/pg_ha_migrations/safe_statements.rb
@@ -26,7 +26,8 @@ module PgHaMigrations::SafeStatements
   end
 
   def safe_add_column(table, column, type, options = {})
-    unless ActiveRecord::Base.connection.postgresql_version >= 110000
+    puts ActiveRecord::Base.connection.postgresql_version
+    unless ActiveRecord::Base.connection.postgresql_version >= 11_00_00
       if options.has_key? :default
         raise PgHaMigrations::UnsafeMigrationError.new(":default is NOT SAFE! Use safe_change_column_default afterwards then backfill the data to prevent locking the table")
       end
--- a/lib/pg_ha_migrations/safe_statements.rb
+++ b/lib/pg_ha_migrations/safe_statements.rb
@@ -26,7 +26,6 @@ module PgHaMigrations::SafeStatements
   end
 
   def safe_add_column(table, column, type, options = {})
-    puts ActiveRecord::Base.connection.postgresql_version
     unless ActiveRecord::Base.connection.postgresql_version >= 11_00_00
       if options.has_key? :default
         raise PgHaMigrations::UnsafeMigrationError.new(":default is NOT SAFE! Use safe_change_column_default afterwards then backfill the data to prevent locking the table")
--- a/lib/pg_ha_migrations/safe_statements.rb
+++ b/lib/pg_ha_migrations/safe_statements.rb
@@ -26,6 +26,10 @@ module PgHaMigrations::SafeStatements
   end
 
   def safe_add_column(table, column, type, options = {})
+    if options.has_key?(:default) && options[:default].include?("(")
+      raise PgHaMigrations::UnsafeMigrationError.new(":default is NOT SAFE! Use safe_change_column_default afterwards then backfill the data to prevent locking the table")
+    end
+
     unless ActiveRecord::Base.connection.postgresql_version >= 11_00_00
       if options.has_key? :default
         raise PgHaMigrations::UnsafeMigrationError.new(":default is NOT SAFE! Use safe_change_column_default afterwards then backfill the data to prevent locking the table")
--- a/lib/pg_ha_migrations/safe_statements.rb
+++ b/lib/pg_ha_migrations/safe_statements.rb
@@ -31,7 +31,7 @@ module PgHaMigrations::SafeStatements
     end
 
     unless ActiveRecord::Base.connection.postgresql_version >= 11_00_00
-      if options.has_key? :default
+      if options.has_key?(:default)
         raise PgHaMigrations::UnsafeMigrationError.new(":default is NOT SAFE! Use safe_change_column_default afterwards then backfill the data to prevent locking the table")
       end
       if options[:null] == false
--- a/spec/safe_statements_spec.rb
+++ b/spec/safe_statements_spec.rb
@@ -512,7 +512,7 @@ RSpec.describe PgHaMigrations::SafeStatements do
         end
 
         describe "safe_add_column" do
-          it "add column without a default" do
+          it "adds column without a default" do
             migration = Class.new(migration_klass) do
               def up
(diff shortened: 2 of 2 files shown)


lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -26,11 +26,17 @@ def safe_add_enum_value(name, value)
   end
 
   def safe_add_column(table, column, type, options = {})
-    if options.has_key? :default
+    if options.has_key?(:default) && options[:default].include?("(")
       raise PgHaMigrations::UnsafeMigrationError.new(":default is NOT SAFE! Use safe_change_column_default afterwards then backfill the data to prevent locking the table")
```
stephenreid: This message could include "with a dynamic value"

Bartlebyy: @jcoleman @JonMR Can you approve this change?

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -26,11 +26,17 @@ def safe_add_enum_value(name, value)
...
     end
-    if options[:null] == false
-      raise PgHaMigrations::UnsafeMigrationError.new(":null => false is NOT SAFE if the table has data! If you _really_ want to do this, use unsafe_make_column_not_nullable")
+
+    unless ActiveRecord::Base.connection.postgresql_version >= 11_00_00
+      if options.has_key? :default
```
jcoleman:
```suggestion
      if options.has_key?(:default)
```

I know it wasn't this way before, but let's correct the style if we're going to change this code.

spec/safe_statements_spec.rb
```diff
@@ -512,24 +512,24 @@ def up
         end
 
         describe "safe_add_column" do
-          it "forbids setting a default" do
+          it "add column without a default" do
```
jcoleman:
```suggestion
          it "adds column without a default" do
```

spec/safe_statements_spec.rb
```diff
@@ -538,17 +538,71 @@ def up
...
+              migration.suppress_messages { migration.migrate(:up) }
+
+              expect(ActiveRecord::Base.connection.columns("foos").detect { |column| column.name == "bar" }.default).to eq("baz")
+
+              ActiveRecord::Base.connection.execute("INSERT INTO foos SELECT FROM (VALUES (1)) t")
+              expect(ActiveRecord::Base.connection.select_value("SELECT bar FROM foos")).to eq("baz")
             end
 
-            migration.suppress_messages { migration.migrate(:up) }
+            it "allows setting null => false" do
```
jcoleman: This still isn't safe with PG11+ right? Adding a `NOT NULL` constraint still requires a full table scan.

spec/safe_statements_spec.rb
```diff
@@ -877,7 +931,7 @@ def up
           end
 
           it "raises if connecting to Postgres 9.1 databases" do
-            allow(ActiveRecord::Base.connection).to receive(:postgresql_version).and_return(90112)
+            allow(ActiveRecord::Base.connection).to receive(:postgresql_version).and_return(9_01_12)
```
jcoleman: This is reasonable cleanup for understandability, but I think it should be a standalone commit.

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -26,11 +26,17 @@ def safe_add_enum_value(name, value)
   end
 
   def safe_add_column(table, column, type, options = {})
-    if options.has_key? :default
+    if options.has_key?(:default) && options[:default].include?("(")
```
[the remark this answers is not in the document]
jcoleman: This is the same problem discussed here by the way: https://github.com/braintree/pg_ha_migrations/pull/37#issuecomment-572728202

spec/safe_statements_spec.rb
jcoleman: Perhaps you were thinking of the feature to convert a `CHECK col IS NOT NULL` constraint into a `NOT NULL` constraint?

JonMR:
Thanks for the PR. Apologies for the slow responses over the holidays.

Can you also change the container launched in the [setup script](https://github.com/braintree/pg_ha_migrations/blob/e798b995d84deb0ab20d1d3054d111c3ec7a538d/bin/setup#L11) to launch a postgres 11 container?

It'd be nice to actually test against specific versions of pg, but that's a bigger item than this PR. 👍

spec/safe_statements_spec.rb
jcoleman: By the way, I have a branch to add more generic constraints, which would lay he groundwork for a function to use the new PG functionality I referenced.

stephenreid: It's not safe for change column, but it's not possible for add column without a default value (because the table is queried for null values)

spec/safe_statements_spec.rb
stephenreid: Moving to its own pr.

lib/pg_ha_migrations/safe_statements.rb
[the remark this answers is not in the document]
stephenreid: Yes, this would trigger in that example; and good point on the lambda compatibility. I'll rethink how to achieve this safely.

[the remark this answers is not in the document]
jcoleman: @JonMR and I discussed this problem pretty heavily on Friday because we were encountering the same difficulty (how to determine if a string represents a SQL functional expression) in e96894de1e8a43b2eee3f72aa641799c35907c40. We ended up punting a bit there because we didn't feel like we'd come up with a way that handles all cases (which is even more important with the new PG11 default-on-new-column features). I slept on it and came up with this rough outline of something that might work:

I'd characterize this as "Best effort but maybe strict enough":
- If column is text or binary, then even a string is fine (since casting can't do anything interesting).
- Otherwise require a scalar object.
- Doesn’t necessarily handle a (custom?) Ruby type and that will be represented by a string in SQL with an explicit type cast. Such a (custom SQL?) type could coerces that string into a function I think, but that seems pretty edge-casey so maybe we’re comfortable with it?
- Possibly overly strict for setting a default on an existing column (e.g., it won’t automatically parse a time stamp string anymore), but we have to be extra careful when allowing a default on a new column since getting this wrong means doing a full table scan. Technically (see previous point) this doesn't inherently handle all theoretical (I don't have any concrete examples) cases here...but pretty close?

We'll discuss a bit more, and if we conclude it seems reasonable, we'll probably try to code this up and commit it for the previously mentioned commit, and then this PR can reuse it.

spec/safe_statements_spec.rb
jcoleman: Oh, I didn't really realize what you were intending at first, and that explanation helped.

In general we're trying to achieve no surprises when a migration runs in a production environment, so I'd tend to consider `safe_add_column :foos, :bar, :text, :null => false` to be unsafe in the sense that it can't be known to be executable (since there could be rows in the table already, which would at least fail _fast_ but still fail).

In our internal process for that case we'd use the `unsafe_add_column` variant and also require that someone from the team owning the table document the row counts in all deployed environments and sign off that the code can't currently write any new rows.

lib/pg_ha_migrations/safe_statements.rb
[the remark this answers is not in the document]
jcoleman: To add a bit more here: I realized that some of the cases we were thinking through only apply in the commit I referenced since they only result in unexpected behavior (like a constant value being set as the default when someone probably intended a runtime expression). That case wouldn't cause any "full table rewrite" problems here.

So I think the approach I detailed -- along with disallowing lambdas entirely in the `safe_add_column` case would solve this.

Am I missing anything?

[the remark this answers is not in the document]
jcoleman: Here's a proof-of-concept on this: #43.

jcoleman:
@stephenreid Now that #43 is merged, I think we have the groundwork necessary to proceed with this. Here's what I think still needs to be done here:

1. We can't allow setting `:null => false` on a new column, because that's a full table scan, and we don't allow full table scans under exclusive locks. Hopefully this particular case will be addressable with future work given changes in PG12+ IIRC that allow converting check constraints into a not null constraint without an exclusive lock being held during the required table scan.
2. Allowing a default to be set on a new column on the appropriate PG versions is good, but we need to make sure there are no expressions. I think the implementation here is basically https://github.com/braintree/pg_ha_migrations/pull/43/files#diff-7031f9f7b6a70dc5abef1a41a89aded1R78-R84 -- we check to make sure that the provided value is either a non-string like type _or_ the column is a string-like type, and we disallow passing lambdas entirely.

jcoleman:
I believe #34 fixes this.

PR #37, opened 2019-11-22
Add appraisal for testing against multiple rails versions

JonMR:
Add appraisal for testing this gem against multiple versions of rails.

This should ultimately help testing Rails 6 in #33


stephenreid:
Hi @JonMR,

We wanted to check on updates of this effort; We have a developer that has used Appraisal before and if it's helpful could help with the effort.

Stephen

JonMR:
Hey @stephenreid, I think the appraisal part is done. The part I'm struggling with is getting one of the specs to pass with rails 5.0 and 5.1. One of the [functions we call](https://github.com/braintree/pg_ha_migrations/blob/e798b995d84deb0ab20d1d3054d111c3ec7a538d/lib/pg_ha_migrations/safe_statements.rb#L50) returns a different value between these versions. I've spent a bit digging through ActiveRecord source and I haven't tracked down yet what the difference is.

The intention of the code and [spec](https://github.com/braintree/pg_ha_migrations/blob/4f7f6219bc8f1205655ef7857d405ae0c7b4db98/spec/safe_statements_spec.rb#L590-L602) is to disallow users from accidentally setting a static value as a default.

Basically, with `safe_change_column_default :foos, :bar, 'NOW()'` the now() doesn't mean execute that function for each row at runtime. It executes it at the time the column default is set. So every row gets a static value (even ones inserted way into the future). If you want to do the right thing in 5.2 at least, you have to do `safe_change_column_default :foos, :bar, -> { 'now()' }`.

If you want to pick this up, it would be best to first check that the behavior detailed in the paragraph above is the same in 5.0 and 5.1. 

If 5.0 and 5.1 don't need the lambda to call the function per row, then we can just skip the check in 5.0 and 5.1. If 5.0 and 5.1 still need the lambda to call the function per row, then we need to understand what's different and find another way to check it.

JonMR:
Ok, we fixed the failing specs on 5.0 and 5.1. I rebased this PR on those changes. Waiting for a passing build, we'll likely merge on Tuesday.

PR #40, opened 2020-01-18
Add methods for managing CHECK constraints

jcoleman:


pg_ha_migrations.gemspec
```diff
@@ -29,7 +29,6 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
-  spec.add_development_dependency "bundler", "~> 1.15"
```
JonMR: Looks like this was included in two PRs.

jcoleman: Oh, yeah. I already merged the other one, so this won't end up in the merged commit.

PR #41, opened 2020-01-20
Use underscore notation for postgres version numbers

stephenreid:
To help human readability in the specs of version numbers.


stephenreid:
Surprises every morning; Thank you!

PR #42, opened 2020-01-21
Cleanup unnecessary output

jcoleman:


lib/pg_ha_migrations/hacks/cleanup_unnecessary_output.rb
```diff
@@ -0,0 +1,29 @@
...
+        connection.adapter_name
+      end
+    end
+  end
+end
+
+patchable_module = [
+ defined?(ActiveRecord::Migration::Compatibility::V5_2) ? ActiveRecord::Migration::Compatibility::V5_2 : nil,
+ defined?(ActiveRecord::Migration::Compatibility::V5_1) ? ActiveRecord::Migration::Compatibility::V5_1 : nil,
+ defined?(ActiveRecord::Migration::Compatibility::V5_0) ? ActiveRecord::Migration::Compatibility::V5_0 : nil,
```
JonMR: Would it make sense to use https://github.com/braintree/pg_ha_migrations/blob/9cb7944e4c79c344416d7947c051afee34803702/lib/pg_ha_migrations/allowed_versions.rb here?

JonMR: Nevermind! I thought you were `map`'ing over these.

PR #45, opened 2020-01-31
Report long transactions without tables locked

jcoleman:
Long-running transactions can block some kinds of DDL (at least from
completing) even if they haven't yet locked any tables. For example,
creating an index concurrently has to wait for all in-flight
transactions to complete before its second phase can run.

Along the way I also modified the reporting to show when a transaction
is idle or now and clarify if the query is currently running or is the
most-recently-run query for an idle transaction.

### DIFF
--- a/lib/pg_ha_migrations/blocking_database_transactions.rb
+++ b/lib/pg_ha_migrations/blocking_database_transactions.rb
@@ -1,12 +1,22 @@
 module PgHaMigrations
   class BlockingDatabaseTransactions
-    LongRunningTransaction = Struct.new(:database, :current_query, :transaction_age, :tables_with_locks) do
+    LongRunningTransaction = Struct.new(:database, :current_query, :state, :transaction_age, :tables_with_locks) do
       def description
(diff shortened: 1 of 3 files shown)


JonMR:
Looks like a few specs failed.

```
  16) PgHaMigrations::SafeStatements ActiveRecord::Migration::Current PgHaMigrations::SafeStatements#safely_acquire_lock_for_table with table_name of type Symbol prints out helpful information when waiting for a lock
      Failure/Error: blocking_transactions.any? { |query| query.tables_with_locks.include?(table) }
      
      NoMethodError:
        undefined method `include?' for nil:NilClass
      # ./lib/pg_ha_migrations/safe_statements.rb:150:in `block in safely_acquire_lock_for_table'
      # ./lib/pg_ha_migrations/safe_statements.rb:150:in `any?'
      # ./lib/pg_ha_migrations/safe_statements.rb:150:in `safely_acquire_lock_for_table'
      # ./spec/safe_statements_spec.rb:1312:in `up'
      # /home/travis/.rvm/gems/ruby-2.4.6/gems/activerecord-5.2.3/lib/active_record/migration.rb:817:in `exec_migration'
      # ./lib/pg_ha_migrations/safe_statements.rb:137:in `exec_migration
```

PR #48, opened 2020-03-11
Bugfix for safe_change_column_default method

lrakhman:
This sets the default value of `safe_change_column_default` to `NULL` if the value passed in is `nil`. Previously, it was setting the value to an empty string.

This is the error encountered while trying to run migrations:
```
-- execute("ALTER TABLE \"archive_gateway_transaction_fees\"\nALTER COLUMN \"id\"\nSET DEFAULT ''\n")
rake aborted!
StandardError: An error has occurred, all later migrations canceled:
PG::InvalidTextRepresentation: ERROR:  invalid input syntax for integer: ""
: ALTER TABLE "archive_gateway_transaction_fees"
ALTER COLUMN "id"
SET DEFAULT ''
```

### DIFF
--- a/lib/pg_ha_migrations/safe_statements.rb
+++ b/lib/pg_ha_migrations/safe_statements.rb
@@ -94,6 +94,10 @@ module PgHaMigrations::SafeStatements
       ERROR
     end
 
+    if default_value.nil?
+      default_value = -> { "NULL" }
+    end
+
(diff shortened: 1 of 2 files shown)


jcoleman:
For history's sake, I interacted out of band with @lrakhman and we were able to track this down to #29 (upgrading to the latest version fixed the issue).

PR #51, opened 2020-06-05
Allow setting a default value in safe_add_column for PostgreSQL 11+

jaronkk:
> From PostgreSQL 11, adding a column with a constant default value no
> longer means that each row of the table needs to be updated when the
> ALTER TABLE statement is executed. Instead, the default value will be
> returned the next time the row is accessed, and applied when the table
> is rewritten, making the ALTER TABLE very fast even on large tables.

https://www.postgresql.org/docs/11/ddl-alter.html#DDL-ALTER-ADDING-A-COLUMN

### DIFF
--- a/lib/pg_ha_migrations/safe_statements.rb
+++ b/lib/pg_ha_migrations/safe_statements.rb
@@ -26,8 +26,14 @@ module PgHaMigrations::SafeStatements
   end
 
   def safe_add_column(table, column, type, options = {})
-    if options.has_key? :default
-      raise PgHaMigrations::UnsafeMigrationError.new(":default is NOT SAFE! Use safe_change_column_default afterwards then backfill the data to prevent locking the table")
+    if options.has_key?(:default)
+      if ActiveRecord::Base.connection.postgresql_version <= 11_00_00
+        raise PgHaMigrations::UnsafeMigrationError.new(":default is NOT SAFE! Use safe_change_column_default afterwards then backfill the data to prevent locking the table")
+      else
+        if options[:default].is_a?(Proc) || (options[:default].is_a?(String) && !([:string, :text, :binary].include?(type.to_sym) || _type_is_enum(type)))
+          raise PgHaMigrations::UnsafeMigrationError.new(":default is not safe if the default value is volatile. Use safe_change_column_default afterwards then backfill the data to prevent locking the table")
+        end
+      end
     end
     if options[:null] == false
       raise PgHaMigrations::UnsafeMigrationError.new(":null => false is NOT SAFE if the table has data! If you _really_ want to do this, use unsafe_make_column_not_nullable")
@@ -199,6 +205,10 @@ module PgHaMigrations::SafeStatements
     expected_adapter = "PostgreSQL"
     actual_adapter = ActiveRecord::Base.connection.adapter_name
     raise PgHaMigrations::UnsupportedAdapter, "This gem only works with the #{expected_adapter} adapter, found #{actual_adapter} instead" unless actual_adapter == expected_adapter
+  end
+
+  def _type_is_enum(type)
+    ActiveRecord::Base.connection.select_values("SELECT typname FROM pg_type JOIN pg_enum ON pg_type.oid = pg_enum.enumtypid").include?(type.to_s)
   end
 
   def migrate(direction)
--- a/lib/pg_ha_migrations/safe_statements.rb
+++ b/lib/pg_ha_migrations/safe_statements.rb
@@ -27,12 +27,10 @@ module PgHaMigrations::SafeStatements
 
   def safe_add_column(table, column, type, options = {})
     if options.has_key?(:default)
-      if ActiveRecord::Base.connection.postgresql_version <= 11_00_00
+      if ActiveRecord::Base.connection.postgresql_version < 11_00_00
         raise PgHaMigrations::UnsafeMigrationError.new(":default is NOT SAFE! Use safe_change_column_default afterwards then backfill the data to prevent locking the table")
-      else
-        if options[:default].is_a?(Proc) || (options[:default].is_a?(String) && !([:string, :text, :binary].include?(type.to_sym) || _type_is_enum(type)))
-          raise PgHaMigrations::UnsafeMigrationError.new(":default is not safe if the default value is volatile. Use safe_change_column_default afterwards then backfill the data to prevent locking the table")
-        end
+      elsif options[:default].is_a?(Proc) || (options[:default].is_a?(String) && !([:string, :text, :binary].include?(type.to_sym) || _type_is_enum(type)))
+        raise PgHaMigrations::UnsafeMigrationError.new(":default is not safe if the default value is volatile. Use safe_change_column_default afterwards then backfill the data to prevent locking the table")
       end
     end
     if options[:null] == false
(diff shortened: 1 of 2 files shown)


jaronkk:
I believe this change would solve #34

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -26,8 +26,14 @@ def safe_add_enum_value(name, value)
   end
 
   def safe_add_column(table, column, type, options = {})
-    if options.has_key? :default
-      raise PgHaMigrations::UnsafeMigrationError.new(":default is NOT SAFE! Use safe_change_column_default afterwards then backfill the data to prevent locking the table")
+    if options.has_key?(:default)
+      if ActiveRecord::Base.connection.postgresql_version <= 11_00_00
```
pgr0ss, line 30: Should this be `<`?

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -201,6 +205,10 @@ def _check_postgres_adapter!
     raise PgHaMigrations::UnsupportedAdapter, "This gem only works with the #{expected_adapter} adapter, found #{actual_adapter} instead" unless actual_adapter == expected_adapter
   end
 
+  def _type_is_enum(type)
```
jcoleman, line 208: Should this also be used in https://github.com/braintree/pg_ha_migrations/pull/43/files#diff-7031f9f7b6a70dc5abef1a41a89aded1R81 ?

jaronkk: You already fixed that in https://github.com/jcoleman/pg_ha_migrations/commit/1395fb4178e8fd474028b22c7a81a129125119b6

`column.sql_type_metadata.type` will return `:enum` for an existing column. The problem here is that there isn't a way (or at least I couldn't find a way) to tell if the type being specified for a new column is an enum without making an extra query.

jcoleman: Heh. Yeah, we do the same thing in the byte packing code to determine if an enum exists.

spec/safe_statements_spec.rb
```diff
@@ -548,11 +548,62 @@ def up
...
+              end.to raise_error PgHaMigrations::UnsafeMigrationError
+            end
+          end
+
+          [:string, :text, :enum, :binary].each do |type|
+            it "allows a default value that looks like an expression for the #{type.inspect} type" do
+              migration = Class.new(migration_klass) do
+                define_method(:up) do
+                  unsafe_create_table :foos
+                  ActiveRecord::Base.connection.execute("INSERT INTO foos SELECT FROM (VALUES (1)) t")
```
jcoleman, line 575:
```suggestion
                  ActiveRecord::Base.connection.execute("INSERT INTO foos(bar) VALUES (1)")
```

spec/safe_statements_spec.rb
```diff
@@ -548,11 +548,62 @@ def up
         end
 
         describe "safe_add_column" do
-          it "forbids setting a default" do
+          it "allows setting a default on Postgres 11+ and forbids it otherwise" do
             migration = Class.new(migration_klass) do
               def up
                 unsafe_create_table :foos
-                safe_add_column :foos, :bar, :text, :default => ""
+                ActiveRecord::Base.connection.execute("INSERT INTO foos SELECT FROM (VALUES (1)) t")
```
jcoleman, line 555:
```suggestion
                ActiveRecord::Base.connection.execute("INSERT INTO foos(bar) VALUES (1)")
```

spec/safe_statements_spec.rb
```diff
@@ -548,11 +548,62 @@ def up
...
+              end
+
+              if ActiveRecord::Base.connection.postgresql_version >= 11_00_00
+                migration.suppress_messages { migration.migrate(:up) }
+
+                # Handle binary columns being transported, but not stored, as hex.
+                expected_value = type == :binary ? "\\x4e4f572829" : "NOW()"
+                expect(ActiveRecord::Base.connection.columns("foos").detect { |column| column.name == "bar" }.default).to eq(expected_value)
+
+                ActiveRecord::Base.connection.execute("INSERT INTO foos SELECT FROM (VALUES (1)) t")
```
jcoleman, line 592:
```suggestion
                ActiveRecord::Base.connection.execute("INSERT INTO foos(bar) VALUES (1)")
```

spec/safe_statements_spec.rb
```diff
@@ -548,11 +548,62 @@ def up
...
+              expect(ActiveRecord::Base.connection.select_value("SELECT bar FROM foos")).to eq("baz")
+            else
+              expect do
+                migration.suppress_messages { migration.migrate(:up) }
+              end.to raise_error PgHaMigrations::UnsafeMigrationError
+            end
+          end
+
+          [:string, :text, :enum, :binary].each do |type|
+            it "allows a default value that looks like an expression for the #{type.inspect} type" do
```
jcoleman, line 571:
```suggestion
            it "allows a default value that looks like an expression for the #{type.inspect} type on Postgres 11+" do
```

spec/safe_statements_spec.rb
```diff
@@ -548,11 +548,62 @@ def up
         end
 
         describe "safe_add_column" do
-          it "forbids setting a default" do
+          it "allows setting a default on Postgres 11+ and forbids it otherwise" do
             migration = Class.new(migration_klass) do
               def up
                 unsafe_create_table :foos
-                safe_add_column :foos, :bar, :text, :default => ""
+                ActiveRecord::Base.connection.execute("INSERT INTO foos(bar) VALUES (1)")
```
jcoleman, line 555:
```suggestion
                ActiveRecord::Base.connection.execute("INSERT INTO foos(id) VALUES (1)")
```

spec/safe_statements_spec.rb
```diff
@@ -548,11 +548,62 @@ def up
...
+              end.to raise_error PgHaMigrations::UnsafeMigrationError
+            end
+          end
+
+          [:string, :text, :enum, :binary].each do |type|
+            it "allows a default value that looks like an expression for the #{type.inspect} type on Postgres 11+" do
+              migration = Class.new(migration_klass) do
+                define_method(:up) do
+                  unsafe_create_table :foos
+                  ActiveRecord::Base.connection.execute("INSERT INTO foos(bar) VALUES (1)")
```
jcoleman, line 575:
```suggestion
                  ActiveRecord::Base.connection.execute("INSERT INTO foos(id) VALUES (1)")
```

spec/safe_statements_spec.rb
```diff
@@ -548,11 +548,62 @@ def up
...
+              end
+
+              if ActiveRecord::Base.connection.postgresql_version >= 11_00_00
+                migration.suppress_messages { migration.migrate(:up) }
+
+                # Handle binary columns being transported, but not stored, as hex.
+                expected_value = type == :binary ? "\\x4e4f572829" : "NOW()"
+                expect(ActiveRecord::Base.connection.columns("foos").detect { |column| column.name == "bar" }.default).to eq(expected_value)
+
+                ActiveRecord::Base.connection.execute("INSERT INTO foos(bar) VALUES (1)")
```
jcoleman, line 592:
```suggestion
                ActiveRecord::Base.connection.execute("INSERT INTO foos(id) VALUES (1)")
```

PR #60, opened 2021-03-05
Enable default password and pg version in bin/setup

celeen:
I hit a few small bumps when running bin/setup and tests. 
Changes:
- The docker container stops if there is no `POSTGRES_PASSWORD` env var set for postgres to use on startup, so this sets it to the same password the specs use to connect to the postgres container
- This updates the default postgres version, and enables the user to specify a different version when running the script

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -2,6 +2,9 @@
 set -euo pipefail
 IFS=$'\n\t'
 set -vx
+export PGPASSWORD="${PGPASSWORD:-postgres}"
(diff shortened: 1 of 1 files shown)


jcoleman (approved):

jcoleman:
I'm re-running CI to see why there are broken jobs.

jcoleman:
Well, it's green now. Something about "push" versus "pull request".

PR #76, opened 2023-04-12
Don't unnecessarily use Exception in lieu of StandardError

jcoleman:
Most exceptions (lowercase) in Ruby should inherit from `StandardError` rather than that class's ancestor of `Exception` (since `Exception` is often not something you should rescue).


bmdavi3:
LGTM

bmdavi3 (approved):

PR #78, opened 2023-05-10
Use docker compose for testing and cleanup build matrix

sp8590:
The main goal of this PR was to get partman installed in the PG image so we can start implementing features and writing tests for partition management. It has snowballed a bit from there 😄 

- Introduce dockerfile with partman and use docker-compose to build and run the container for local testing and CI
- Remove EOL versions of ruby, rails, and postgres from the build matrix / appraisals
- Tighten the rails dependency in the gemspec to `>= 6.1` and  `< 7.1`
- Cleanup conditional logic in tests to accommodate the new build matrix


bmdavi3 (approved):
LGTM

rkrage:
Closing as this was addressed in:
- https://github.com/braintree/pg_ha_migrations/pull/79
- https://github.com/braintree/pg_ha_migrations/pull/80

PR #79, opened 2023-05-18
Remove EOL versions in build matrix / gemspec

rkrage:
Rails 6.0 isn't _technically_ EOL yet, but it will be in two weeks. We still have some conditional logic when running against older postgres versions (which seems valuable). Now that PG 9.6 and 10 are no longer included in the build matrix I've updated tests that exercise that logic to stub the call to `ActiveRecord::Base.connection.postgresql_version`.


lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -273,15 +273,9 @@ def safe_create_partitioned_table(table, key:, type:, infer_primary_key: nil, **
...
-          next unless col.primary_key
+        next unless col.options[:primary_key]
 
-          col.primary_key = false
-        else
-          next unless col.options[:primary_key]
-
-          col.options[:primary_key] = false
-        end
+        col.options[:primary_key] = false
```
rkrage, line 278: This conditional logic is no longer needed, as the API for accessing the primary key is consistent across Rails 6.1 and 7.0

jcoleman (approved):

PR #80, opened 2023-05-18
Build custom image with partman for testing

rkrage:
We're planning to introduce some features around partition management that utilize pg_partman. It doesn't look like there's a well maintained public docker image for postgres that includes partman, so let's use docker-compose for local / CI testing and build our own!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,21 +22,12 @@ jobs:
     env: # $BUNDLE_GEMFILE must be set at the job level, so it is set for all steps
       BUNDLE_GEMFILE: gemfiles/${{ matrix.gemfile }}.gemfile
       ImageOS: ubuntu20
-    services:
-      postgresql:
-        image: postgres:${{ matrix.pg }}
-        env:
-          POSTGRES_PASSWORD: postgres
-        # Set health checks to wait until postgres has started
-        options: >-
-          --health-cmd pg_isready
-          --health-interval 10s
-          --health-timeout 5s
-          --health-retries 5
-        ports:
-          - 5432:5432
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
+      - name: Build postgres image and start the container
+        run: docker-compose up -d --build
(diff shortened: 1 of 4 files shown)


spec/spec_helper.rb
```diff
@@ -17,6 +17,13 @@
     server_version = ActiveRecord::Base.connection.select_value("SHOW server_version")
     puts "DEBUG: Connecting to Postgres server version #{server_version}"
 
+    ActiveRecord::Base.connection.execute("DROP EXTENSION IF EXISTS pg_partman CASCADE")
+
+    # Drop parent partition tables first to automatically drop children
+    ActiveRecord::Base.connection.select_values("SELECT c.relname FROM pg_class c JOIN pg_partitioned_table p on c.oid = p.partrelid").each do |table|
+      ActiveRecord::Base.connection.execute("DROP TABLE #{table} CASCADE")
+    end
+
```
rkrage, line 26: All of this code isn't _strictly_ necessary yet as we haven't added any tests that create the partman extension / child partitions. We can definitely remove it from this PR if folks feel strongly, but it shouldn't affect existing tests in any way

spec/spec_helper.rb
```diff
@@ -26,8 +33,14 @@
   end
 
   config.after(:each) do
-    # ActiveRecord::Base.connection.tables does not include partitioned tables in Rails 5.1
-    ActiveRecord::Base.connection.select_values("SELECT tablename FROM pg_tables WHERE schemaname = 'public'").each do |table|
+    ActiveRecord::Base.connection.execute("DROP EXTENSION IF EXISTS pg_partman CASCADE")
```
jcoleman, line 36: Earlier I'd noted we should probably bump this to a PR that needs it, but when we do that it strikes me that we should probably also extract this code so that we don't have to maintain it two different places in the same file.

rkrage: Yup that's probably a good idea. We'll include that as a separate commit in the next partman PR

jcoleman (approved):

PR #81, opened 2023-05-31
Methods for maintaining partitions with partman

rkrage:
Would like to get some feedback on this approach before we go ahead and write the docs. There are likely some additional partman features we would want, but figured this was a good place to start.

Update: the docs have been written

### DIFF
--- a/README.md
+++ b/README.md
@@ -325,8 +325,7 @@ safe_partman_create_parent :table,
 
 #### unsafe\_partman\_create\_parent
 
-We have chosen to flag the use of `retention` and `retention_keep_table` as an unsafe operation.
-While we recognize that these options are useful, we think they fit in the same category as `drop_table` and `rename_table`, and are therefore unsafe from an application perspective.
+As mentioned above, we have chosen to flag the use of `retention` and `retention_keep_table` as an unsafe operation.
 If you wish to define these options, you must use this method.
 
 ```ruby
--- a/README.md
+++ b/README.md
@@ -286,7 +286,7 @@ These options are delegated to the `unsafe_partman_update_config` method to upda
 
 - `infinite_time_partitions`. Partman defaults this to `false` but we default to `true`
 - `inherit_privileges`. Partman defaults this to `false` but we default to `true`
-- `retention`. Default: Partman defaults this to `null`
+- `retention`. Partman defaults this to `null`
 - `retention_keep_table`. Partman defaults this to `true`
 
 ```ruby
--- a/README.md
+++ b/README.md
@@ -226,18 +226,18 @@ Safely create a new partitioned table using [declaritive partitioning](https://w
 
 ```ruby
 # list partitioned table using single column as partition key
-safe_create_partitioned_table :table, type: :list, key: :example_column do |t|
+safe_create_partitioned_table :table, type: :list, partition_key: :example_column do |t|
   t.text :example_column, null: false
 end
 
 # range partitioned table using multiple columns as partition key
-safe_create_partitioned_table :table, type: :range, key: [:example_column_a, :example_column_b] do |t|
+safe_create_partitioned_table :table, type: :range, partition_key: [:example_column_a, :example_column_b] do |t|
   t.integer :example_column_a, null: false
   t.integer :example_column_b, null: false
 end
 
 # hash partitioned table using expression as partition key
-safe_create_partitioned_table :table, :type: :hash, key: ->{ "(example_column::date)" } do |t|
+safe_create_partitioned_table :table, :type: :hash, partition_key: ->{ "(example_column::date)" } do |t|
   t.datetime :example_column, null: false
 end
 ```
@@ -245,7 +245,7 @@ end
 The identifier column type is `bigserial` by default. This can be overridden, as you would in `safe_create_table`, by setting the `id` argument:
 
 ```ruby
-safe_create_partitioned_table :table, id: :serial, type: :range, key: :example_column do |t|
+safe_create_partitioned_table :table, id: :serial, type: :range, partition_key: :example_column do |t|
   t.date :example_column, null: false
 end
 ```
@@ -254,12 +254,12 @@ In PostgreSQL 11+, primary key constraints are supported on partitioned tables g
 
 ```ruby
 # primary key will be (id, example_column)
-safe_create_partitioned_table :table, type: :range, key: :example_column do |t|
+safe_create_partitioned_table :table, type: :range, partition_key: :example_column do |t|
   t.date :example_column, null: false
 end
 
 # primary key will not be created
-safe_create_partitioned_table :table, type: :range, key: :example_column, infer_primary_key: false do |t|
+safe_create_partitioned_table :table, type: :range, partition_key: :example_column, infer_primary_key: false do |t|
   t.date :example_column, null: false
 end
 ```
@@ -274,7 +274,7 @@ The first (and only) positional argument maps to `p_parent_table` in the `create
 
 The rest are keyword args with the following mappings:
 
-- `key` -> `p_control`. Required: `true`
+- `partition_key` -> `p_control`. Required: `true`
 - `interval` -> `p_interval`. Required: `true`
 - `template_table` -> `p_template_table`. Required: `false`. Partman will create a template table if not defined.
 - `premake` -> `p_premake`. Required: `false`. Partman defaults to `4`.
@@ -293,17 +293,17 @@ These options are delegated to the `unsafe_partman_update_config` method to upda
 With only the required args:
 
 ```ruby
-safe_create_partitioned_table :table, type: :range, key: :created_at do |t|
+safe_create_partitioned_table :table, type: :range, partition_key: :created_at do |t|
   t.timestamps null: false
 end
 
-safe_partman_create_parent :table, key: :created_at, interval: "weekly"
+safe_partman_create_parent :table, partition_key: :created_at, interval: "weekly"
 ```
 
 With custom overrides:
 
 ```ruby
-safe_create_partitioned_table :table, type: :range, key: :created_at do |t|
+safe_create_partitioned_table :table, type: :range, partition_key: :created_at do |t|
   t.timestamps null: false
   t.text :some_column
 end
@@ -314,11 +314,11 @@ safe_create_table :table_template, id: false do |t|
 end
 
 safe_partman_create_parent :table,
-  key: :created_at,
+  partition_key: :created_at,
   interval: "weekly",
   template_table: :table_template,
   premake: 10,
-  start_partition: (Time.current + 1.month).to_fs(:db),
+  start_partition: Time.current + 1.month,
   infinite_time_partitions: false,
   inherit_privileges: false
 ```
@@ -330,12 +330,12 @@ While we recognize that these options are useful, we think they fit in the same 
 If you wish to define these options, you must use this method.
 
 ```ruby
-safe_create_partitioned_table :table, type: :range, key: :created_at do |t|
+safe_create_partitioned_table :table, type: :range, partition_key: :created_at do |t|
   t.timestamps null: false
 end
 
 unsafe_partman_create_parent :table,
-  key: :created_at,
+  partition_key: :created_at,
   interval: "weekly",
   retention: "60 days",
   retention_keep_table: false
--- a/README.md
+++ b/README.md
@@ -270,7 +270,7 @@ Safely configure a partitioned table to be managed by [pg\_partman](https://gith
 
 This method calls the [create\_parent](https://github.com/pgpartman/pg_partman/blob/master/doc/pg_partman.md#creation-functions) partman function with some reasonable defaults and a subset of user-defined overrides.
 
-The first (and only) positional arg maps to `p_parent_table` in the `create_parent` function.
+The first (and only) positional argument maps to `p_parent_table` in the `create_parent` function.
 
 The rest are keyword args with the following mappings:
 
@@ -294,7 +294,7 @@ With only the required args:
 
 ```ruby
 safe_create_partitioned_table :table, type: :range, key: :created_at do |t|
-  t.timestamps :null => false
+  t.timestamps null: false
 end
 
 safe_partman_create_parent :table, key: :created_at, interval: "weekly"
@@ -304,13 +304,13 @@ With custom overrides:
 
 ```ruby
 safe_create_partitioned_table :table, type: :range, key: :created_at do |t|
-  t.timestamps :null => false
+  t.timestamps null: false
   t.text :some_column
 end
 
 # Partman will reference the template table to create unique indexes on child tables
 safe_create_table :table_template, id: false do |t|
-  t.text :some_column, index: { unique: true }
+  t.text :some_column, index: {unique: true}
 end
 
 safe_partman_create_parent :table,
@@ -331,7 +331,7 @@ If you wish to define these options, you must use this method.
 
 ```ruby
 safe_create_partitioned_table :table, type: :range, key: :created_at do |t|
-  t.timestamps :null => false
+  t.timestamps null: false
 end
 
 unsafe_partman_create_parent :table,
@@ -366,7 +366,7 @@ safe_partman_update_config :table,
 
 #### unsafe\_partman\_update\_config
 
-As mentioned previously, we have chosen to flag the use of `retention` and `retention_keep_table` as an unsafe operation.
+As with creating a partman parent table, we have chosen to flag the use of `retention` and `retention_keep_table` as an unsafe operation.
 If you wish to define these options, you must use this method.
 
 ```ruby
--- a/README.md
+++ b/README.md
@@ -276,6 +276,7 @@ The rest are keyword args with the following mappings:
 
 - `key` -> `p_control`. Required: `true`
 - `interval` -> `p_interval`. Required: `true`
+- `template_table` -> `p_template_table`. Required: `false`. Partman will create a template table if not defined.
 - `premake` -> `p_premake`. Required: `false`. Partman defaults to `4`.
 - `start_partition` -> `p_start_partition`. Required: `false`. Partman defaults to the current timestamp.
 
@@ -304,11 +305,18 @@ With custom overrides:
 ```ruby
 safe_create_partitioned_table :table, type: :range, key: :created_at do |t|
   t.timestamps :null => false
+  t.text :some_column
+end
+
+# Partman will reference the template table to create unique indexes on child tables
+safe_create_table :table_template, id: false do |t|
+  t.text :some_column, index: { unique: true }
 end
 
 safe_partman_create_parent :table,
   key: :created_at,
   interval: "weekly",
+  template_table: :table_template,
   premake: 10,
   start_partition: (Time.current + 1.month).to_fs(:db),
   infinite_time_partitions: false,
--- a/README.md
+++ b/README.md
@@ -289,6 +289,18 @@ These options are delegated to the `unsafe_partman_update_config` method to upda
 - `retention`. Partman defaults this to `null`
 - `retention_keep_table`. Partman defaults this to `true`
 
+With only the required args:
+
+```ruby
+safe_create_partitioned_table :table, type: :range, key: :created_at do |t|
+  t.timestamps :null => false
+end
+
+safe_partman_create_parent :table, key: :created_at, interval: "weekly"
+```
+
+With custom overrides:
+
 ```ruby
 safe_create_partitioned_table :table, type: :range, key: :created_at do |t|
   t.timestamps :null => false
--- a/README.md
+++ b/README.md
@@ -325,7 +325,8 @@ safe_partman_create_parent :table,
 
 #### unsafe\_partman\_create\_parent
 
-As mentioned above, we have chosen to flag the use of `retention` and `retention_keep_table` as an unsafe operation.
+We have chosen to flag the use of `retention` and `retention_keep_table` as an unsafe operation.
+While we recognize that these options are useful, we think they fit in the same category as `drop_table` and `rename_table`, and are therefore unsafe from an application perspective.
 If you wish to define these options, you must use this method.
 
 ```ruby
@@ -365,8 +366,7 @@ safe_partman_update_config :table,
 
 #### unsafe\_partman\_update\_config
 
-We have chosen to flag the use of `retention` and `retention_keep_table` as an unsafe operation.
-While we recognize that these options are useful, we think they fit in the same category as `drop_table` and `rename_table`, and are therefore unsafe from an application perspective.
+As mentioned previously, we have chosen to flag the use of `retention` and `retention_keep_table` as an unsafe operation.
 If you wish to define these options, you must use this method.
 
 ```ruby
--- a/README.md
+++ b/README.md
@@ -264,6 +264,106 @@ safe_create_partitioned_table :table, type: :range, key: :example_column, infer_
 end
 ```
 
+#### safe\_partman\_create\_parent
+
+Safely configure a partitioned table to be managed by [pg\_partman](https://github.com/pgpartman/pg_partman).
+
+This method calls the [create\_parent](https://github.com/pgpartman/pg_partman/blob/master/doc/pg_partman.md#creation-functions) partman function with some reasonable defaults and a subset of user-defined overrides.
+
+The first (and only) positional arg maps to `p_parent_table` in the `create_parent` function.
+
+The rest are keyword args with the following mappings:
+
+- `key` -> `p_control`. Required: `true`
+- `interval` -> `p_interval`. Required: `true`
+- `premake` -> `p_premake`. Required: `false`. Partman defaults to `4`.
+- `start_partition` -> `p_start_partition`. Required: `false`. Partman defaults to the current timestamp.
+
+Note that we have chosen to require PostgreSQL 11+ and hardcode `p_type` to `native` for simplicity, as previous PostgreSQL versions are end-of-life.
+
+Additionally, this method allows you to configure a subset of attributes on the record stored in the [part\_config](https://github.com/pgpartman/pg_partman/blob/master/doc/pg_partman.md#tables) table.
+These options are delegated to the `unsafe_partman_update_config` method to update the record:
+
+- `infinite_time_partitions`. Partman defaults this to `false` but we default to `true`
+- `inherit_privileges`. Partman defaults this to `false` but we default to `true`
+- `retention`. Default: Partman defaults this to `null`
+- `retention_keep_table`. Partman defaults this to `true`
+
+```ruby
+safe_create_partitioned_table :table, type: :range, key: :created_at do |t|
+  t.timestamps :null => false
+end
+
+safe_partman_create_parent :table,
+  key: :created_at,
+  interval: "weekly",
+  premake: 10,
+  start_partition: (Time.current + 1.month).to_fs(:db),
+  infinite_time_partitions: false,
+  inherit_privileges: false
+```
+
+#### unsafe\_partman\_create\_parent
+
+We have chosen to flag the use of `retention` and `retention_keep_table` as an unsafe operation.
+While we recognize that these options are useful, we think they fit in the same category as `drop_table` and `rename_table`, and are therefore unsafe from an application perspective.
+If you wish to define these options, you must use this method.
+
+```ruby
+safe_create_partitioned_table :table, type: :range, key: :created_at do |t|
+  t.timestamps :null => false
+end
+
+unsafe_partman_create_parent :table,
+  key: :created_at,
+  interval: "weekly",
+  retention: "60 days",
+  retention_keep_table: false
+```
+
+#### safe\_partman\_update\_config
+
+There are some partitioning options that cannot be set in the call to `create_parent` and are only available in the `part_config` table.
+As mentioned previously, you can specify these args in the call to `safe_partman_create_parent` or `unsafe_partman_create_parent` which will be delegated to this method.
+Calling this method directly will be useful if you need to modify your partitioned table after the fact.
+
+Allowed keyword args:
+
+- `infinite_time_partitions`
+- `inherit_privileges`
+- `premake`
+- `retention`
+- `retention_keep_table`
+
+Note that we detect if the value of `inherit_privileges` is changing and will automatically call `safe_partman_reapply_privileges` to ensure permissions are propagated to existing child partitions.
+
+```ruby
+safe_partman_update_config :table,
+  infinite_time_partitions: false,
+  inherit_privileges: false,
+  premake: 10
+```
+
+#### unsafe\_partman\_update\_config
+
+We have chosen to flag the use of `retention` and `retention_keep_table` as an unsafe operation.
+While we recognize that these options are useful, we think they fit in the same category as `drop_table` and `rename_table`, and are therefore unsafe from an application perspective.
+If you wish to define these options, you must use this method.
+
+```ruby
+unsafe_partman_update_config :table,
+  retention: "60 days",
+  retention_keep_table: false
+```
+
+#### safe\_partman\_reapply\_privileges
+
+If your partitioned table is configured with `inherit_privileges` set to `true`, use this method after granting new roles / privileges on the parent table to ensure permissions are propagated to existing child partitions.
+
+```ruby
+safe_partman_reapply_privileges :table
+```
+
 ### Utilities
 
 #### safely\_acquire\_lock\_for\_table
--- a/lib/pg_ha_migrations/partman_config.rb
+++ b/lib/pg_ha_migrations/partman_config.rb
@@ -1,6 +1,10 @@
+# This is an internal class that is not meant to be used directly
 class PgHaMigrations::PartmanConfig < ActiveRecord::Base
   self.primary_key = :parent_table
 
+  # This method is called by unsafe_partman_update_config to set the fully
+  # qualified table name, as partman is often installed in a schema that
+  # is not included the application's search path
   def self.schema=(schema)
     self.table_name = "#{schema}.part_config"
   end
--- a/lib/pg_ha_migrations/partman_config.rb
+++ b/lib/pg_ha_migrations/partman_config.rb
@@ -1,0 +1,7 @@
+class PgHaMigrations::PartmanConfig < ActiveRecord::Base
+  self.primary_key = :parent_table
+
+  def self.schema=(schema)
+    self.table_name = "#{schema}.part_config"
+  end
+end
--- a/lib/pg_ha_migrations/safe_statements.rb
+++ b/lib/pg_ha_migrations/safe_statements.rb
@@ -433,6 +433,7 @@ module PgHaMigrations::SafeStatements
       FROM pg_tables
       WHERE tablename = '#{identifiers.last}' AND schemaname = #{schema_conditional}
       ORDER BY array_position(current_schemas(false), schemaname)
+      LIMIT 1
     SQL
 
     raise PgHaMigrations::InvalidMigrationError, "Could not find table #{table}" unless schema.present?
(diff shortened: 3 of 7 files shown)


spec/safe_statements_spec.rb
```diff
@@ -2110,6 +2110,555 @@ def up
...
+            it "raises error when parent table does not exist and fully qualified name provided" do
+              migration = Class.new(migration_klass) do
+                def up
+                  safe_partman_create_parent "public.foos3", key: :created_at, interval: "monthly"
+                end
+              end
+
+              expect do
+                migration.suppress_messages { migration.migrate(:up) }
+              end.to raise_error(ActiveRecord::StatementInvalid, /Please create parent table first: public.foos3/)
```
rkrage, line 2364: In a handful of places, we're simply bubbling up an error that comes from partman. Do we want to capture these and reraise as an `PgHaMigrations::InvalidMigrationError`?

lib/pg_ha_migrations/partman_config.rb
```diff
@@ -0,0 +1,7 @@
+class PgHaMigrations::PartmanConfig < ActiveRecord::Base
+  self.primary_key = :parent_table
+
+  def self.schema=(schema)
+    self.table_name = "#{schema}.part_config"
```
jcoleman, line 5: Should we default this to `partman.part_config`?

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -286,6 +300,76 @@ def safe_create_partitioned_table(table, key:, type:, infer_primary_key: nil, **
     end
   end
 
+  def safe_partman_create_parent(table, key:, interval:, **options)
```
jcoleman, line 303: Should this have explicit keyword args instead of options? Then we get the validation of valid option names for free.

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -286,6 +300,76 @@ def safe_create_partitioned_table(table, key:, type:, infer_primary_key: nil, **
...
+    raise ArgumentError, "Expected <key> to be present" unless key.present?
+    raise ArgumentError, "Expected <interval> to be present" unless interval.present?
+
+    invalid_options = options.keys - PARTMAN_CREATE_PARENT_OPTIONS
+
+    raise ArgumentError, "Unrecognized optional argument(s): #{invalid_options}" unless invalid_options.empty?
+
+    if ActiveRecord::Base.connection.postgresql_version < 10_00_00
+      raise PgHaMigrations::InvalidMigrationError, "Native partitioning not supported on Postgres databases before version 10"
```
jcoleman, line 312: I think we should explicitly deprecate (i.e., remove) < PG11 sooner rather than later.

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -286,6 +300,76 @@ def safe_create_partitioned_table(table, key:, type:, infer_primary_key: nil, **
...
+    options[:template_table] = _fully_qualified_name(options[:template_table]) if options[:template_table].present?
+
+    options = options.merge(
+      parent_table: _fully_qualified_name(table),
+      control: key,
+      type: "native",
+      interval: interval,
+    ).compact
+
+    create_parent_sql = options.map { |k, v| "p_#{k} := '#{v}'" }.join(", ")
```
jcoleman, line 324: I don't love manually munging SQL, but we do it elsewhere. Nonetheless, we should probably use the PG gem's string escaping code at least.

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -286,6 +300,76 @@ def safe_create_partitioned_table(table, key:, type:, infer_primary_key: nil, **
...
+
+    connection.execute("SELECT #{_quoted_partman_schema}.create_parent(#{create_parent_sql})")
+  end
+
+  def safe_partman_update_config(table, **options)
```
jcoleman, line 329: I think this could benefit from explicit keyword args also.

spec/spec_helper.rb
```diff
@@ -1,6 +1,7 @@
 require "bundler/setup"
 require "pg_ha_migrations"
 require "db-query-matchers"
+require "pry-byebug"
```
jcoleman, line 4: This should be in the `.pryrc` file; if we don't have one we should add it.

jcoleman (commented):
I didn't look at tests yet, but I wanted to leave initial thoughts.

lib/pg_ha_migrations/partman_config.rb
rkrage: That doesn't seem necessary, as this class isn't meant to be used directly. In `safe_partman_update_config` we always discover the schema and call this method before interacting with the class

lib/pg_ha_migrations/safe_statements.rb
rkrage: That is true, but it comes at the cost of a really long method signature and would require us to manually build the hash of arguments here: https://github.com/braintree/pg_ha_migrations/blob/a0275cda749115a94122a0e42bed5c915e105cdf/lib/pg_ha_migrations/safe_statements.rb#L317-L322

lib/pg_ha_migrations/safe_statements.rb
rkrage: I'm not sure it makes sense to completely disallow < PG11 for the entire library, but maybe it makes sense for new features (especially this feature since partman recommends PG11+)

lib/pg_ha_migrations/safe_statements.rb
rkrage: I think this one actually has a stronger argument against using keyword args. Since this is updating a row in the db, the presence of an argument with a nil value actually has meaning, and you wouldn't be able to distinguish that with a method signature like:

```ruby
def foo(a: nil, b: nil)
```

We probably wouldn't want to null out columns `b` if the method is called with `foo(a: "some value")`. And if we filter out options with nil values, we wouldn't be able to set a column to null when called with `foo(a: nil)`

spec/spec_helper.rb
rkrage: So we do have a `.pryrc` file with logic like this: https://github.com/braintree/pg_ha_migrations/blob/92ec97c9ffca4785b48ca84bb71afb914e827776/.pryrc#L1-L6

Googling around, I don't see a ton of examples that are requiring `pry-byebug` in this file

lib/pg_ha_migrations/safe_statements.rb
jcoleman: Oh, hmm, that's a good point.

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -286,6 +299,96 @@ def safe_create_partitioned_table(table, key:, type:, infer_primary_key: nil, **
...
+
+    safe_partman_update_config(fully_qualified_table_name, **update_config_options)
+  end
+
+  def safe_partman_update_config(table, **options)
+    invalid_options = options.keys - PARTMAN_UPDATE_CONFIG_OPTIONS
+
+    raise ArgumentError, "Unrecognized argument(s): #{invalid_options}" unless invalid_options.empty?
+
+    PgHaMigrations::PartmanConfig.schema = _quoted_partman_schema
```
jcoleman, line 347: This feels pretty weird IMO.

rkrage: I agree, but I do think it's a pretty nice user experience to auto-discover the schema. The alternative here would maybe be to define a config value for the partman schema that defaults to `partman` but can be overridden:

```ruby
PgHaMigrations.configure do |config|
  config.partman_schema = "public"
end
```

lib/pg_ha_migrations/safe_statements.rb
jcoleman: Yeah, I'm not saying there are no downsides, but I think it also offers real benefit to the callers of the code, because there's self-documentation in the signature.

lib/pg_ha_migrations/safe_statements.rb
jcoleman: > I'm not sure it makes sense to completely disallow < PG11 for the entire library

I'm sure :D

PG11 itself will be EOL this fall; anything else has already been EOL for more than half a year, and that after a 5 year cycle. I think it's completely natural to follow the PostgreSQL projections EOL policy.

lib/pg_ha_migrations/safe_statements.rb
rkrage: Because this method is also passing along args to `unsafe_partman_update_config`, I can sort of make the same argument that I made below. However, the only column that is actually nullable is `retention` and that value is null by default, so there's no difference between calling this method with `retention: nil` and just omitting the argument altogether. So that's my long winded way of saying I implemented your proposed change and pushed up a commit 😆

lib/pg_ha_migrations/safe_statements.rb
rkrage: Fair enough, but let's tackle that in another PR. In the meantime I pushed up a commit to change this guard from < 10 to < 11

spec/spec_helper.rb
```diff
@@ -1,7 +1,7 @@
 require "bundler/setup"
 require "pg_ha_migrations"
 require "db-query-matchers"
-require "pry-byebug"
+require "pry"
```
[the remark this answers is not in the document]
rkrage: We talked offline about this a little bit. I think it's convenient to have this for folks not using our vim setup, such that you don't need to do this every time you want to pry:

```ruby
require "pry"
binding.pry
```

But I also don't feel strongly, so I'm happy to do whatever here

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -286,6 +294,130 @@ def safe_create_partitioned_table(table, key:, type:, infer_primary_key: nil, **
...
+    raise PgHaMigrations::InvalidMigrationError, "Found #{table} in multiple schemas: #{schemas}" if schemas.size > 1
+
+    _validate_partman_identifier(schemas.first)
+
+    "#{schemas.first}.#{table}"
+  end
+
+  def _validate_partman_identifier(identifier)
+    if identifier.to_s !~ /^[a-z_][a-z_\d]*$/
+      raise PgHaMigrations::InvalidMigrationError, "Partman requires schema / table names to be lowercase with underscores"
```
rkrage, line 417: You previously made a comment about escaping the values we pass to `create_parent` and that made me realize we weren't sanitizing table names in any way. Since we're calling partman functions and passing the table name as a string and aren't just referencing it as an identifier in a normal query, we can't use things like `quote_table_name`. Now, I also tried using things like `quote` and `escape_string`, but that generated all sorts of errors in the internal partman functions. What I've gathered is partman has very limited support for tables with non-standard names, so I think it seems valuable to enforce that in our code to avoid strange errors from partman

README.md
```diff
@@ -264,6 +264,125 @@ safe_create_partitioned_table :table, type: :range, key: :example_column, infer_
...
+- `key` -> `p_control`. Required: `true`
+- `interval` -> `p_interval`. Required: `true`
+- `template_table` -> `p_template_table`. Required: `false`. Partman will create a template table if not defined.
+- `premake` -> `p_premake`. Required: `false`. Partman defaults to `4`.
+- `start_partition` -> `p_start_partition`. Required: `false`. Partman defaults to the current timestamp.
+
+Note that we have chosen to require PostgreSQL 11+ and hardcode `p_type` to `native` for simplicity, as previous PostgreSQL versions are end-of-life.
+
+Additionally, this method allows you to configure a subset of attributes on the record stored in the [part\_config](https://github.com/pgpartman/pg_partman/blob/master/doc/pg_partman.md#tables) table.
+These options are delegated to the `unsafe_partman_update_config` method to update the record:
```
rkrage, line 286: > Some projects attempt to hide complexity by having code determine the intent and magically do the right series of operations. But we (and by extension this gem) take the approach that it's better to understand exactly what the database is doing so that (particularly long running) operations are not a surprise during your deploy cycle.

Do we think this violates this principle?

README.md
```diff
@@ -264,6 +264,125 @@ safe_create_partitioned_table :table, type: :range, key: :example_column, infer_
...
+
+Allowed keyword args:
+
+- `infinite_time_partitions`
+- `inherit_privileges`
+- `premake`
+- `retention`
+- `retention_keep_table`
+
+Note that we detect if the value of `inherit_privileges` is changing and will automatically call `safe_partman_reapply_privileges` to ensure permissions are propagated to existing child partitions.
```
rkrage, line 358: Same comment as above. Is this too magical?

spec/safe_statements_spec.rb
jcoleman: I'd probably prefer validation first.

lib/pg_ha_migrations/safe_statements.rb
jcoleman: Yeah, that is kinda convenient.

spec/spec_helper.rb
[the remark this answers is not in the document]
jcoleman: I don't really have a strong opinion here either; I have a minor preference for not having it, but it's fine either way.

README.md
jcoleman: No, I think this is just eliminating a sharp edge from partman.

README.md
```diff
@@ -264,6 +264,125 @@ safe_create_partitioned_table :table, type: :range, key: :example_column, infer_
...
+```ruby
+safe_partman_update_config :table,
+  infinite_time_partitions: false,
+  inherit_privileges: false,
+  premake: 10
+```
+
+#### unsafe\_partman\_update\_config
+
+As mentioned previously, we have chosen to flag the use of `retention` and `retention_keep_table` as an unsafe operation.
```
jcoleman, line 369:
```suggestion
As with creating a partman parent table, we have chosen to flag the use of `retention` and `retention_keep_table` as an unsafe operation.
```

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -1,5 +1,19 @@
...
+    start_partition
+    epoch
+    template_table
+    jobmon
+  ]
+
+  PARTMAN_UPDATE_CONFIG_OPTIONS = %i[
+    infinite_time_partitions
+    inherit_privileges
+  ]
```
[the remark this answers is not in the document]
jcoleman: Have we effectively done this now?

lib/pg_ha_migrations/partman_config.rb
jcoleman: I think we should add a comment here explaining what's going on. I think it makes sense now, but it's still pretty weird, and I'd like signposts in the future.

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -286,6 +294,130 @@ def safe_create_partitioned_table(table, key:, type:, infer_primary_key: nil, **
...
+      start_partition: start_partition,
+    }.compact
+
+    if template_table.present?
+      create_parent_options[:template_table] = _fully_qualified_partman_name(template_table)
+    end
+
+    create_parent_sql = create_parent_options.map { |k, v| "p_#{k} := #{connection.quote(v)}" }.join(", ")
+
+    unsafe_execute("SELECT #{_quoted_partman_schema}.create_parent(#{create_parent_sql})")
```
jcoleman, line 341: We should probably wrap this in a `say_with_time` block like we used to have to do when we implemented our own check constraint methods: https://github.com/braintree/pg_ha_migrations/pull/84/files#diff-e246218fba61b707b474c288f842dc0bf0c83436253bf2b49d05216d222f034cL181

(and the other commands too)

lib/pg_ha_migrations/safe_statements.rb
jcoleman: That all makes sense, except that probably it's OK with uppercase? Lowercase is, of course, the Rails convention, so I don't feel strongly about supporting that, but still curious.

spec/safe_statements_spec.rb
```diff
@@ -2110,6 +2110,762 @@ def up
...
+            migration.safe_partman_create_parent(:foos3, arg1: "foo", arg2: "bar")
+          end
+        end
+
+        describe "unsafe_partman_create_parent" do
+          context "when extension not installed" do
+            it "raises error" do
+              migration = Class.new(migration_klass) do
+                def up
+                  safe_create_partitioned_table :foos3, type: :range, key: :created_at do |t|
```
jcoleman, line 2154: It might be nice to extract this to a `before(:each)`, which will also separate it from the actual code under test.

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -286,6 +294,130 @@ def safe_create_partitioned_table(table, key:, type:, infer_primary_key: nil, **
...
+        ON pg_namespace.oid = pg_extension.extnamespace
+      WHERE pg_extension.extname = 'pg_partman'
+    SQL
+
+    raise PgHaMigrations::InvalidMigrationError, "The pg_partman extension is not installed" unless schema.present?
+
+    connection.quote_schema_name(schema)
+  end
+
+  def _fully_qualified_partman_name(table)
```
jcoleman, line 396: This name could probably improve; maybe just `_fully_qualified_table_name`? Or, if it's only used for the partman configuration, then `_fully_qualified_table_name_for_partman_config`?

spec/safe_statements_spec.rb
```diff
@@ -2110,6 +2110,762 @@ def up
...
+                    t.text :text_column
+                  end
+
+                  unsafe_partman_create_parent :foos3, key: :created_at, interval: "monthly"
+                end
+              end
+
+              expect do
+                migration.suppress_messages { migration.migrate(:up) }
+              end.to raise_error(ActiveRecord::StatementInvalid, /you must have created the given parent table as ranged \(not list\) partitioned already/)
```
jcoleman, line 2488:
```suggestion
              end.to raise_error(ActiveRecord::StatementInvalid, /you must have created the given parent table as range \(not list\) partitioned already/)
```

spec/spec_helper.rb
```diff
@@ -61,3 +51,33 @@
 ActiveRecord::Tasks::DatabaseTasks.drop_current
 ActiveRecord::Tasks::DatabaseTasks.create_current
 ActiveRecord::Base.establish_connection(config)
+
+class DatabaseHelper
```
jcoleman, line 55:
```suggestion
module DatabaseHelper
```

spec/spec_helper.rb
```diff
@@ -61,3 +51,33 @@
...
+  class << self
```
jcoleman, line 56: Personal preference here, but I'd probably just use `def self.drop_objects` style.

README.md
```diff
@@ -264,6 +264,125 @@ safe_create_partitioned_table :table, type: :range, key: :example_column, infer_
...
+
+Safely configure a partitioned table to be managed by [pg\_partman](https://github.com/pgpartman/pg_partman).
+
+This method calls the [create\_parent](https://github.com/pgpartman/pg_partman/blob/master/doc/pg_partman.md#creation-functions) partman function with some reasonable defaults and a subset of user-defined overrides.
+
+The first (and only) positional arg maps to `p_parent_table` in the `create_parent` function.
+
+The rest are keyword args with the following mappings:
+
+- `key` -> `p_control`. Required: `true`
```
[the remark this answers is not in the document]
rkrage: I used `key` here to be consistent with the method signature of `safe_create_partitioned_table`. However, we haven't released a new version since that method was added so we _could_ change it in both places without breaking backwards compatibility if you feel strongly

README.md
```diff
@@ -264,6 +264,125 @@ safe_create_partitioned_table :table, type: :range, key: :example_column, infer_
...
+safe_create_table :table_template, id: false do |t|
+  t.text :some_column, index: { unique: true }
+end
+
+safe_partman_create_parent :table,
+  key: :created_at,
+  interval: "weekly",
+  template_table: :table_template,
+  premake: 10,
+  start_partition: (Time.current + 1.month).to_fs(:db),
```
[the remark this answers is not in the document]
rkrage: Hmm maybe... So partman does support using an integer (representing epoch time) as the partition key, _but_ for simplicity our code does not support that (you would need to set `p_epoch`). So maybe we can assume this value is either a `Time` object or a string representing a `Time` object and parse it accordingly?

lib/pg_ha_migrations/partman_config.rb
rkrage: Comments added

lib/pg_ha_migrations/safe_statements.rb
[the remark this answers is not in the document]
rkrage: Yes, we had a design discussion about it. I think the summary is that we should start simple and only allow config options that _we_ need and flag the use of certain options as unsafe (like `retention` and `retention_keep_table`). If this gains traction in the community we can add on to it over time

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -229,8 +237,8 @@ def unsafe_remove_constraint(table, name:)
   def safe_create_partitioned_table(table, key:, type:, infer_primary_key: nil, **options, &block)
     raise ArgumentError, "Expected <key> to be present" unless key.present?
 
-    unless VALID_PARTITION_TYPES.include?(type)
-      raise ArgumentError, "Expected <type> to be symbol in #{VALID_PARTITION_TYPES}"
+    unless PARTITION_TYPES.include?(type)
+      raise ArgumentError, "Expected <type> to be symbol in #{PARTITION_TYPES}"
```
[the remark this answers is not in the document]
rkrage: This code isn't actually relevant to this PR, but seems like a reasonable suggestion. There's a diff here because I renamed `VALID_PARTITION_TYPES` to `PARTITION_TYPES` to be more consistent with the other constant that was added in this PR.

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -286,6 +294,130 @@ def safe_create_partitioned_table(table, key:, type:, infer_primary_key: nil, **
     end
   end
 
+  def safe_partman_create_parent(table, **options)
+    if options[:retention].present? || options[:retention_keep_table] == false
```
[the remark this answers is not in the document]
rkrage: So you're suggesting that we have keyword args, but exclude `retention` and `retention_keep_table` such that it results in an argument error when called with those options? I think the existing error is _much_ more helpful and consistent with other methods in this library

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -286,6 +294,130 @@ def safe_create_partitioned_table(table, key:, type:, infer_primary_key: nil, **
...
+
+    connection.quote_schema_name(schema)
+  end
+
+  def _fully_qualified_partman_name(table)
+    table.to_s.split(".").each { |part| _validate_partman_identifier(part) }
+
+    return table if table.to_s.include?(".")
+
+    schemas = connection.select_values(<<~SQL)
```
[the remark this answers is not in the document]
rkrage: I believe the original code tried to grab the table from the _first_ schema. We had a design discussion about this and came to the conclusion that ambiguous tables (existing in multiple schemas) could cause some confusion. However, this doesn't align with other Rails methods like `create_table`, so maybe it's a good idea to be consistent?

The original code: https://github.com/braintree/pg_ha_migrations/pull/81/commits/4b7a418fa0522e630be34d0ddd2f4cac4536504a#diff-e246218fba61b707b474c288f842dc0bf0c83436253bf2b49d05216d222f034cR358-R371

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -286,6 +294,130 @@ def safe_create_partitioned_table(table, key:, type:, infer_primary_key: nil, **
...
+      FROM pg_tables
+      WHERE tablename = '#{table}' AND schemaname = ANY (current_schemas(false))
+    SQL
+
+    raise PgHaMigrations::InvalidMigrationError, "Could not find #{table} in search path" unless schemas.present?
+    raise PgHaMigrations::InvalidMigrationError, "Found #{table} in multiple schemas: #{schemas}" if schemas.size > 1
+
+    _validate_partman_identifier(schemas.first)
+
+    "#{schemas.first}.#{table}"
```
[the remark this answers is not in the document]
rkrage: I added a comment here about why we _don't_ need quoting

lib/pg_ha_migrations/safe_statements.rb
rkrage: In the absence of quoting (because partman doesn't like it), the use of uppercase letters would cause some confusion as I believe they would just get converted to lowercase. That would also be inconsistent with other Rails methods so I'd prefer to completely disallow uppercase

spec/safe_statements_spec.rb
rkrage: It is a little difficult though, because partman does _a ton_ of validation, so we'd essentially be rewriting a chunk of partman in ruby. That being said though, we _do_ have (and need) some validation when doing table discovery. If you simply provide a table name that doesn't exist, you'll get a pg ha migrations error. If you provide a fully qualified table name that doesn't exist, you'll get statement invalid error from partman. So it's probably at least worth making those types of errors consistent. I'll look into that

lib/pg_ha_migrations/safe_statements.rb
rkrage: Done

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -286,6 +294,130 @@ def safe_create_partitioned_table(table, key:, type:, infer_primary_key: nil, **
...
+
+    inherit_privileges_changed = config.inherit_privileges_changed?
+
+    config.save!
+
+    safe_partman_reapply_privileges(table) if inherit_privileges_changed
+  end
+
+  def safe_partman_reapply_privileges(table)
+    unsafe_execute("SELECT #{_quoted_partman_schema}.reapply_privileges('#{_fully_qualified_partman_name(table)}')")
```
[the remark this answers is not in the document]
rkrage: It looks like it does a series of grants and revokes, so there will be some locks taken out on internal tables, but nothing dangerous on the partition tables themselves: https://github.com/pgpartman/pg_partman/blob/master/sql/functions/apply_privileges.sql

spec/safe_statements_spec.rb
rkrage: I do think that makes sense, but it would require a pretty significant refactor to remove stuff like this from the body of _all_ tests (since there are some negative tests that rely on the parent table _not_ being present)

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -286,6 +294,130 @@ def safe_create_partitioned_table(table, key:, type:, infer_primary_key: nil, **
...
+    raise ArgumentError, "Expected <interval> to be present" unless interval.present?
+
+    if ActiveRecord::Base.connection.postgresql_version < 11_00_00
+      raise PgHaMigrations::InvalidMigrationError, "Native partitioning with partman not supported on Postgres databases before version 11"
+    end
+
+    fully_qualified_table_name = _fully_qualified_partman_name(table)
+
+    create_parent_options = {
+      parent_table: fully_qualified_table_name,
```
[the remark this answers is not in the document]
rkrage: We can't do exactly what you described since I don't think we ever want to set the template table as the parent table, but I think I know what you're getting at and there are some changes in the most recent commit to make this cleaner

README.md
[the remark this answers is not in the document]
jcoleman: I think that would be valuable. I did later notice that method also, and it's slightly less meaningful there, but I think it will improve readability in both cases (potentially a good amount).

README.md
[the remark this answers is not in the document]
jcoleman: I think we should just required it to be a `Time` or `Date` or `DateTime` instance.

lib/pg_ha_migrations/safe_statements.rb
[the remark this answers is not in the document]
jcoleman: That's a good argument, and I'd thought on that, OTOH that's a _runtime_ error, while the kwargs error would be preventing the code from being valid at all. Though...it's Ruby, so...maybe no editor is going to check that anyway.

Either way I think we should make the args consistent.

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -286,6 +294,141 @@ def safe_create_partitioned_table(table, key:, type:, infer_primary_key: nil, **
...
+      control: key,
+      type: "native",
+      interval: interval,
+      premake: premake,
+      start_partition: start_partition,
+    }.compact
+
+    create_parent_sql = create_parent_options.map { |k, v| "p_#{k} := #{connection.quote(v)}" }.join(", ")
+
+    log_message = "partman_create_parent(#{table}, " \
```
jcoleman, line 336: Should we use `<<~EOS` style for ease of reading?

lib/pg_ha_migrations/safe_statements.rb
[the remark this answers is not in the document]
jcoleman: Verified with:

In one session:
```
test=# create table bogus(i int);
CREATE TABLE
test=# begin;
BEGIN
test=# create user someuser;
CREATE ROLE
test=# grant select on bogus to someuser;
GRANT
test=#
```
and in another session that changes locks from:
```
test=# select * from pg_locks;
  locktype  | database | relation | page | tuple | virtualxid | transactionid | classid | objid | objsubid | virtualtransaction |  pid  |      mode       | granted | fastpath
------------+----------+----------+------+-------+------------+---------------+---------+-------+----------+--------------------+-------+-----------------+---------+----------
 relation   |  3328774 |    11645 |      |       |            |               |         |       |          | 3/2554606          | 27175 | AccessShareLock | t       | t
 virtualxid |          |          |      |       | 3/2554606  |               |         |       |          | 3/2554606          | 27175 | ExclusiveLock   | t       | t
 virtualxid |          |          |      |       | 4/863422   |               |         |       |          | 4/863422           | 27167 | ExclusiveLock   | t       | t
(3 rows)
```
to:
```
test=# select * from pg_locks;
   locktype    | database | relation | page | tuple | virtualxid | transactionid | classid |  objid   | objsubid | virtualtransaction |  pid  |       mode       | granted | fastpath
---------------+----------+----------+------+-------+------------+---------------+---------+----------+----------+--------------------+-------+------------------+---------+----------
 relation      |  3328774 |    11645 |      |       |            |               |         |          |          | 3/2554607          | 27175 | AccessShareLock  | t       | t
 virtualxid    |          |          |      |       | 3/2554607  |               |         |          |          | 3/2554607          | 27175 | ExclusiveLock    | t       | t
 virtualxid    |          |          |      |       | 4/863422   |               |         |          |          | 4/863422           | 27167 | ExclusiveLock    | t       | t
 transactionid |          |          |      |       |            |       7706875 |         |          |          | 4/863422           | 27167 | ExclusiveLock    | t       | f
 relation      |        0 |     1260 |      |       |            |               |         |          |          | 4/863422           | 27167 | RowExclusiveLock | t       | f
 object        |        0 |          |      |       |            |               |    1260 | 12081725 |        0 | 4/863422           | 27167 | AccessShareLock  | t       | f
(6 rows)
```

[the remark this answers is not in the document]
jcoleman: So it only acquires locks on an internal relation.

spec/safe_statements_spec.rb
jcoleman: Hmm, well, my bigger point is that having the code under test not be mixed with the code not under test is meaningful with these block based expectations.

lib/pg_ha_migrations/safe_statements.rb
jcoleman: Sure!

spec/safe_statements_spec.rb
jcoleman: Is this one of the things we should consider validating?

lib/pg_ha_migrations/safe_statements.rb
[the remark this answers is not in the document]
rkrage: From a functionality perspective, the only thing we gain by having kwargs is that the argument validation happens before the check for `retention` and `retention_keep_table`. Without kwargs the argument validation happens in the unsafe version of the method and I don't think that's a big deal. I'm just hesitant to do this because delegating to another method when kwargs are present becomes very verbose (and ugly IMO):

```ruby
def foo(a, b:, c:, d:, e:)
  # do some validation then delegate to bar
  bar(a, b: b, c: c, d: d, e: e)
end
```

I know that kwargs provide better documentation for the method signature, but if anyone who is vaguely familiar with ruby is looking at this code, I think it's pretty clear that this method is just delegating and they should reference the unsafe version of the method to see the actual method signature

lib/pg_ha_migrations/safe_statements.rb
rkrage: I mean maybe, but that introduces newline characters so we'd have to split on `\n` and then join on `, `

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -280,10 +288,171 @@ def safe_create_partitioned_table(table, key:, type:, infer_primary_key: nil, **
...
+        begin
+          formatted_start_partition = start_partition.send(meth, :db)
+        rescue NoMethodError, ArgumentError
+        end
+
+        break if formatted_start_partition.present?
+      end
+
+      if !formatted_start_partition.present?
+        raise PgHaMigrations::InvalidMigrationError, "Expected <start_partition> to be Date, Time, or DateTime object but received #{start_partition.class}"
```
rkrage, line 338: The goal of this `if` block was to allow duck-typed date / time objects while also using the appropriate method (`to_fs` vs `to_s`) for different rails versions, but I don't really love this code.

It would probably be more readable if we just check if `start_partition.is_a?` date / time object upfront and then call the appropriate method with a conditional (`if rails < 7 do this else do that`)

spec/safe_statements_spec.rb
rkrage: I did fix this _particular_ error so that we're now raising a `PgHaMigrations::InvalidMigrationError` for consistency

spec/safe_statements_spec.rb
rkrage: This goes back to my comment here: https://github.com/braintree/pg_ha_migrations/pull/81#discussion_r1286185575

If we do that, where does it end?

lib/pg_ha_migrations/safe_statements.rb
rkrage: Yeah I was trying to be way too clever here. I rewrote this and pushed a change

spec/safe_statements_spec.rb
jcoleman: Fair enough. I feel like there probably should be some core set, but I also haven't looked in detail at the underlying code.

lib/pg_ha_migrations/safe_statements.rb
jcoleman: So the heredoc form can also have the `\` right before the line breaks, FWIW.

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -280,10 +288,165 @@ def safe_create_partitioned_table(table, key:, type:, infer_primary_key: nil, **
...
+      "'#{identifiers.first}'"
+    else
+      "ANY (current_schemas(false))"
+    end
+
+    schema = connection.select_value(<<~SQL)
+      SELECT schemaname
+      FROM pg_tables
+      WHERE tablename = '#{identifiers.last}' AND schemaname = #{schema_conditional}
+      ORDER BY array_position(current_schemas(false), schemaname)
```
jcoleman, line 435:
```suggestion
      ORDER BY array_position(current_schemas(false), schemaname)
      LIMIT 1
```

jcoleman (approved):
Please squash into useful commits before merging (or if you think it all belongs in a single commit, use GH's squash and merge).

PR #84, opened 2023-07-06
Use Rails built-in {add,remove}_check_constraint

jcoleman:
When we initially added this I'm not sure these methods existed. Since they do now we should build on top of them as well as apply our standard method for preventing using the native methods directly without an `unsafe_` prefix.

### DIFF
--- a/lib/pg_ha_migrations/safe_statements.rb
+++ b/lib/pg_ha_migrations/safe_statements.rb
@@ -170,20 +170,6 @@ module PgHaMigrations::SafeStatements
     unsafe_add_check_constraint(table, expression, name: name, validate: false)
   end
 
-  def unsafe_add_check_constraint(table, expression, name:, validate: true)
-    raise ArgumentError, "Expected <name> to be present" unless name.present?
-
-    quoted_table_name = connection.quote_table_name(table)
-    quoted_constraint_name = connection.quote_table_name(name)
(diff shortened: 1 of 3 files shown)


lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -173,14 +173,8 @@ def safe_add_unvalidated_check_constraint(table, expression, name:)
   def unsafe_add_check_constraint(table, expression, name:, validate: true)
```
rkrage, line 173: Can we just straight up get rid of this method definition?

rkrage: For other methods, like `drop_table`, I'm not seeing an explicitly defined `unsafe_*` method where we're just delegating to `super`

jcoleman: So this is different from the standard implementation I think:
```
raise ArgumentError, "Expected <name> to be present" unless name.present?
```
But it's not clear whether we want to keep it or not.

jcoleman: I think I'm going to make an executive decision and say we don't need it (we don't require it for e.g. indexes).

I'll update this PR in a bit.

rkrage (approved):

Issue #39, opened 2020-01-18
Safe Lock Acquisition Should Specifically Target Different Lock Modes


jcoleman:
Currently `safely_acquire_lock_for_table` always acquires an `AccessExclusiveLock`, which is generally safe in the sense that:
- It's always sufficient for the operation since it's the highest lock type for a table.
- Most of the places we use it need that high of a lock level anyway.

However, it has a few disadvantages:
- There are some existing places where we don't need that high of a lock level (albeit the operations are still fast, so it hasn't seemed like a big deal).
- We could use it in places where we explicitly do not need a high level lock.

If we change this, we'd need t make sure that nesting calls (or even calls in the same transaction -- even though we discourage that) can't upgrade locks since that's could result in a deadlock.

PR #85, opened 2023-08-11
Method for adding partitioned indexes concurrently

rkrage:
Following the recommendations in the postgres docs: https://www.postgresql.org/docs/current/ddl-partitioning.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,7 +223,7 @@ Notes:
 
 - This method does not support sub-partitioning.
 - This method runs multiple DDL statements non-transactionally.
-  - creating / attaching an index on a child table could fail, resulting in an error and an invalid index on the parent table
+  - Creating or attaching an index on a child table could fail. In such cases an exception will be raised, and an `INVALID` index will be left on the parent table.
 
 #### safe\_add\_unvalidated\_check\_constraint
 
--- a/README.md
+++ b/README.md
@@ -222,7 +222,7 @@ safe_add_concurrent_partitioned_index :partitioned_table, [:column1, :column2], 
 Notes:
 
 - this method does not support sub-partitioning
-- this method runs multiple DDL statements non-transactionally
+- This method runs multiple DDL statements non-transactionally.
   - creating / attaching an index on a child table could fail, resulting in an error and an invalid index on the parent table
 
 #### safe\_add\_unvalidated\_check\_constraint
--- a/README.md
+++ b/README.md
@@ -204,14 +204,14 @@ Add an index to a natively partitioned table concurrently, as described in the [
 safe_add_concurrent_partitioned_index :partitioned_table, :column
 ```
 
-Add a composite index with custom naming.
+Add a composite index with custom naming using the `hash` index type.
 
 ```ruby
 # Assuming this table has partitions child1 and child2, the following indexes will be created:
 #   - partitioned_table_index_suffix
 #   - child1_index_suffix (attached to partitioned_table_index_suffix)
 #   - child2_index_suffix (attached to partitioned_table_index_suffix)
-safe_add_concurrent_partitioned_index :partitioned_table, [:column1, :column2], name_suffix: "index_suffix"
+safe_add_concurrent_partitioned_index :partitioned_table, [:column1, :column2], name_suffix: "index_suffix", using: :hash
 ```
 
 Notes:
--- a/README.md
+++ b/README.md
@@ -76,6 +76,11 @@ The following functionality is currently unsupported:
 - Generators
 - schema.rb
 
+Compatibility notes:
+
+- While some features may work with older versions, this gem is currently tested against PostgreSQL 11+
+- There is a [bug](https://github.com/rails/rails/pull/41490) in early versions of Rails 6.1 when using `algorithm: :concurrently`. To add / remove indexes concurrently, please upgrade to at least Rails 6.1.4.
+
 #### safe\_create\_table
 
 Safely creates a new table.
--- a/README.md
+++ b/README.md
@@ -201,17 +201,17 @@ Add an index to a natively partitioned table concurrently, as described in the [
 #   - partitioned_table_column_idx
 #   - child1_column_idx (attached to partitioned_table_column_idx)
 #   - child2_column_idx (attached to partitioned_table_column_idx)
-safe_add_concurrent_partitioned_index :partitioned_table, :column
+safe_add_concurrent_partitioned_index :partitioned_table, :column, name_suffix: "column_idx"
 ```
 
-Add a composite index with custom naming using the `hash` index type.
+Add a composite index using the `hash` index type.
 
 ```ruby
 # Assuming this table has partitions child1 and child2, the following indexes will be created:
-#   - partitioned_table_index_suffix
-#   - child1_index_suffix (attached to partitioned_table_index_suffix)
-#   - child2_index_suffix (attached to partitioned_table_index_suffix)
-safe_add_concurrent_partitioned_index :partitioned_table, [:column1, :column2], name_suffix: "index_suffix", using: :hash
+#   - partitioned_table_column1_column2_idx
+#   - child1_column1_column2_idx (attached to partitioned_table_column1_column2_idx)
+#   - child2_column1_column2_idx (attached to partitioned_table_column1_column2_idx)
+safe_add_concurrent_partitioned_index :partitioned_table, [:column1, :column2], name_suffix: "column1_column2_idx", using: :hash
 ```
 
 Notes:
--- a/README.md
+++ b/README.md
@@ -213,10 +213,10 @@ Add a composite index using the `hash` index type with custom name suffix.
 
 ```ruby
 # Assuming this table has partitions child1 and child2, the following indexes will be created:
-#   - index_partitioned_table_custom_name
-#   - index_child1_custom_name (attached to index_partitioned_table_custom_name)
-#   - index_child2_custom_name (attached to index_partitioned_table_custom_name)
-safe_add_concurrent_partitioned_index :partitioned_table, [:column1, :column2], name_suffix: "custom_name", using: :hash
+#   - custom_name_idx
+#   - index_child1_on_column1_column2 (attached to custom_name_idx)
+#   - index_child2_on_column1_column2 (attached to custom_name_idx)
+safe_add_concurrent_partitioned_index :partitioned_table, [:column1, :column2], name: "custom_name_idx", using: :hash
 ```
 
 Notes:
--- a/README.md
+++ b/README.md
@@ -195,7 +195,7 @@ safe_remove_concurrent_index :table, :name => :index_name
 
 #### safe\_add\_concurrent\_partitioned\_index
 
-Add an index to a natively partitioned table concurrently, as described in the [table partitioning docs](https://www.postgresql.org/docs/current/ddl-partitioning.html).
+Add an index to a natively partitioned table concurrently, as described in the [table partitioning docs](https://www.postgresql.org/docs/current/ddl-partitioning.html):
 
 > To avoid long lock times, it is possible to use `CREATE INDEX ON ONLY` the partitioned table; such an index is marked invalid, and the partitions do not get the index applied automatically.
 > The indexes on partitions can be created individually using `CONCURRENTLY`, and then attached to the index on the parent using `ALTER INDEX .. ATTACH PARTITION`.
--- a/README.md
+++ b/README.md
@@ -188,6 +188,34 @@ Safely remove an index. Migrations that contain this statement must also include
 safe_remove_concurrent_index :table, :name => :index_name
 ```
 
+#### safe\_add\_concurrent\_partitioned\_index
+
+Add an index to a natively partitioned table concurrently, as described in the [table partitioning docs](https://www.postgresql.org/docs/current/ddl-partitioning.html).
+
+> To avoid long lock times, it is possible to use `CREATE INDEX ON ONLY` the partitioned table; such an index is marked invalid, and the partitions do not get the index applied automatically.
+> The indexes on partitions can be created individually using `CONCURRENTLY`, and then attached to the index on the parent using `ALTER INDEX .. ATTACH PARTITION`.
+> Once indexes for all partitions are attached to the parent index, the parent index is marked valid automatically.
+
+```ruby
+# Assuming this table has partitions child1 and child2, the following indexes will be created:
+#   - partitioned_table_column_idx
+#   - child1_column_idx (attached to partitioned_table_column_idx)
+#   - child2_column_idx (attached to partitioned_table_column_idx)
+safe_add_concurrent_partitioned_index :partitioned_table, :column
+```
+
+Add a composite index with custom naming.
+
+```ruby
+# Assuming this table has partitions child1 and child2, the following indexes will be created:
+#   - partitioned_table_index_suffix
+#   - child1_index_suffix (attached to partitioned_table_index_suffix)
+#   - child2_index_suffix (attached to partitioned_table_index_suffix)
+safe_add_concurrent_partitioned_index :partitioned_table, [:column1, :column2], name_suffix: "index_suffix"
+```
+
+Note: this method does not support sub-partitioning.
+
 #### safe\_add\_unvalidated\_check\_constraint
 
 Safely add a `CHECK` constraint. The constraint will not be immediately validated on existing rows to avoid a full table scan while holding an exclusive lock. After adding the constraint, you'll need to use `safe_validate_check_constraint` to validate existing rows.
--- a/README.md
+++ b/README.md
@@ -214,7 +214,11 @@ Add a composite index with custom naming.
 safe_add_concurrent_partitioned_index :partitioned_table, [:column1, :column2], name_suffix: "index_suffix"
 ```
 
-Note: this method does not support sub-partitioning.
+Notes:
+
+- this method does not support sub-partitioning
+- this method runs multiple DDL statements non-transactionally
+  - creating / attaching an index on a child table could fail, resulting in an error and an invalid index on the parent table
 
 #### safe\_add\_unvalidated\_check\_constraint
 
--- a/README.md
+++ b/README.md
@@ -203,20 +203,20 @@ Add an index to a natively partitioned table concurrently, as described in the [
 
 ```ruby
 # Assuming this table has partitions child1 and child2, the following indexes will be created:
-#   - partitioned_table_column_idx
-#   - child1_column_idx (attached to partitioned_table_column_idx)
-#   - child2_column_idx (attached to partitioned_table_column_idx)
-safe_add_concurrent_partitioned_index :partitioned_table, :column, name_suffix: "column_idx"
+#   - index_partitioned_table_on_column
+#   - index_child1_on_column (attached to index_partitioned_table_on_column)
+#   - index_child2_on_column (attached to index_partitioned_table_on_column)
+safe_add_concurrent_partitioned_index :partitioned_table, :column
 ```
 
-Add a composite index using the `hash` index type.
+Add a composite index using the `hash` index type with custom name suffix.
 
 ```ruby
 # Assuming this table has partitions child1 and child2, the following indexes will be created:
-#   - partitioned_table_column1_column2_idx
-#   - child1_column1_column2_idx (attached to partitioned_table_column1_column2_idx)
-#   - child2_column1_column2_idx (attached to partitioned_table_column1_column2_idx)
-safe_add_concurrent_partitioned_index :partitioned_table, [:column1, :column2], name_suffix: "column1_column2_idx", using: :hash
+#   - index_partitioned_table_custom_name
+#   - index_child1_custom_name (attached to index_partitioned_table_custom_name)
+#   - index_child2_custom_name (attached to index_partitioned_table_custom_name)
+safe_add_concurrent_partitioned_index :partitioned_table, [:column1, :column2], name_suffix: "custom_name", using: :hash
 ```
 
 Notes:
--- a/README.md
+++ b/README.md
@@ -221,7 +221,7 @@ safe_add_concurrent_partitioned_index :partitioned_table, [:column1, :column2], 
 
 Notes:
 
-- this method does not support sub-partitioning
+- This method does not support sub-partitioning.
 - This method runs multiple DDL statements non-transactionally.
   - creating / attaching an index on a child table could fail, resulting in an error and an invalid index on the parent table
 
--- a/lib/pg_ha_migrations/hacks/add_index_on_only.rb
+++ b/lib/pg_ha_migrations/hacks/add_index_on_only.rb
@@ -14,13 +14,10 @@ module PgHaMigrations
         if o.algorithm == "ONLY"
           o.algorithm = nil
 
-          original_query = super
+          quoted_index = quote_column_name(o.index.name)
+          quoted_table = quote_table_name(o.index.table)
 
-          if original_query.gsub(" ON ").count > 1
-            raise PgHaMigrations::InvalidMigrationError, "Found multiple occurrences of \"ON\" in query string for index #{o.index.name.inspect}; cannot safely replace with \"ON ONLY\""
-          end
-
-          original_query.sub(" ON ", " ON ONLY ")
+          super.sub("#{quoted_index} ON #{quoted_table}", "#{quoted_index} ON ONLY #{quoted_table}")
         else
           super
         end
--- a/lib/pg_ha_migrations/hacks/add_index_on_only.rb
+++ b/lib/pg_ha_migrations/hacks/add_index_on_only.rb
@@ -20,7 +20,7 @@ module PgHaMigrations
             raise PgHaMigrations::InvalidMigrationError, "Found multiple occurrences of \"ON\" in query string for index #{o.index.name.inspect}; cannot safely replace with \"ON ONLY\""
           end
 
-          super.sub(" ON ", " ON ONLY ")
+          original_query.sub(" ON ", " ON ONLY ")
         else
           super
         end
--- a/lib/pg_ha_migrations/hacks/add_index_on_only.rb
+++ b/lib/pg_ha_migrations/hacks/add_index_on_only.rb
@@ -1,0 +1,33 @@
+require "active_record/connection_adapters/postgresql_adapter"
+require "active_record/connection_adapters/postgresql/schema_creation"
+
+module PgHaMigrations
+  module ActiveRecordHacks
+    module IndexAlgorithms
+      def index_algorithms
+        super.merge(only: "ONLY")
+      end
+    end
+
+    module CreateIndexDefinition
+      def visit_CreateIndexDefinition(o)
+        if o.algorithm == "ONLY"
+          o.algorithm = nil
+
+          original_query = super
+
+          if original_query.gsub(" ON ").count > 1
+            raise PgHaMigrations::InvalidMigrationError, "Found multiple occurrences of \"ON\" in query string for index #{o.index.name.inspect}; cannot safely replace with \"ON ONLY\""
+          end
+
+          super.sub(" ON ", " ON ONLY ")
+        else
+          super
+        end
+      end
+    end
+  end
+end
+
+ActiveRecord::ConnectionAdapters::PostgreSQLAdapter.prepend(PgHaMigrations::ActiveRecordHacks::IndexAlgorithms)
+ActiveRecord::ConnectionAdapters::PostgreSQL::SchemaCreation.prepend(PgHaMigrations::ActiveRecordHacks::CreateIndexDefinition)
--- a/lib/pg_ha_migrations/safe_statements.rb
+++ b/lib/pg_ha_migrations/safe_statements.rb
@@ -177,18 +177,7 @@ module PgHaMigrations::SafeStatements
 
     schema, table = _schema_and_table_for(table)
 
-    is_partitioned = select_value(<<~SQL)
-      SELECT EXISTS (
-        SELECT 1
-        FROM pg_partitioned_table, pg_class, pg_namespace
-        WHERE pg_class.oid = pg_partitioned_table.partrelid
-          AND pg_class.relnamespace = pg_namespace.oid
-          AND pg_class.relname = #{quote(table)}
-          AND pg_namespace.nspname = #{quote(schema)}
-      )
-    SQL
-
-    raise PgHaMigrations::InvalidMigrationError, "Table #{table} is not a partitioned table" unless is_partitioned
+    raise PgHaMigrations::InvalidMigrationError, "Table #{table} is not a partitioned table" unless _partitioned_table?(schema, table)
 
     name_suffix = "#{Array.wrap(columns).join("_")}_idx" unless name_suffix.present?
 
@@ -201,6 +190,10 @@ module PgHaMigrations::SafeStatements
       WHERE parent.relname = #{quote(table)}
         AND pg_namespace.nspname = #{quote(schema)}
     SQL
+
+    child_tables.each do |child_table|
+      raise PgHaMigrations::InvalidMigrationError, "Partitioned table #{table} contains sub-partitions" if _partitioned_table?(schema, child_table)
+    end
 
     parent_index = "#{table}_#{name_suffix}"
 
@@ -507,6 +500,19 @@ module PgHaMigrations::SafeStatements
     else
       !is_valid.nil?
     end
+  end
+
+  def _partitioned_table?(schema, table)
+    select_value(<<~SQL)
+      SELECT EXISTS (
+        SELECT 1
+        FROM pg_partitioned_table, pg_class, pg_namespace
+        WHERE pg_class.oid = pg_partitioned_table.partrelid
+          AND pg_class.relnamespace = pg_namespace.oid
+          AND pg_class.relname = #{quote(table)}
+          AND pg_namespace.nspname = #{quote(schema)}
+      )
+    SQL
   end
 
   def _per_migration_caller
--- a/lib/pg_ha_migrations/safe_statements.rb
+++ b/lib/pg_ha_migrations/safe_statements.rb
@@ -1,5 +1,4 @@
 module PgHaMigrations::SafeStatements
-  MAX_INDEX_NAME_SIZE = 63 # bytes
   PARTITION_TYPES = %i[range list hash]
 
   PARTMAN_UPDATE_CONFIG_OPTIONS = %i[
@@ -174,14 +173,13 @@ module PgHaMigrations::SafeStatements
   def safe_add_concurrent_partitioned_index(
     table,
     columns,
-    name_suffix:,
+    name_suffix: nil,
     if_not_exists: nil,
     using: nil,
     unique: nil,
     where: nil,
     comment: nil
   )
-    raise ArgumentError, "Expected <name_suffix> to be present" unless name_suffix.present?
 
     if ActiveRecord::Base.connection.postgresql_version < 11_00_00
       raise PgHaMigrations::InvalidMigrationError, "Concurrent partitioned index creation not supported on Postgres databases before version 11"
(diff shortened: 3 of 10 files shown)


lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -170,6 +170,65 @@ def safe_remove_concurrent_index(table, options={})
...
+      child_index = "#{child_table}_#{name_suffix}"
+
+      next if if_not_exists && _index_exists?(schema, child_index)
+
+      safe_add_concurrent_index("#{quote_schema_name(schema)}.#{child_table}", columns, name: child_index)
+
+      unsafe_execute(<<~SQL)
+        ALTER INDEX #{quote_schema_name(schema)}.#{quote_column_name(parent_index)}
+        ATTACH PARTITION #{quote_schema_name(schema)}.#{quote_column_name(child_index)}
+      SQL
```
rkrage, line 224: This should be relatively quick but it takes out an access exclusive lock. Should we wrap this in a call to `adjust_lock_timeout` or `adjust_statement_timeout`?

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -170,6 +170,65 @@ def safe_remove_concurrent_index(table, options={})
     unsafe_remove_index(table, **options.merge(:algorithm => :concurrently))
   end
 
+  def safe_add_concurrent_partitioned_index(table, columns, name_suffix: nil, if_not_exists: false)
```
rkrage, line 173: Do we want to add `using:` here to support index types besides btree?

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -170,6 +170,65 @@ def safe_remove_concurrent_index(table, options={})
...
+
+    child_tables = select_values(<<~SQL)
+      SELECT child.relname
+      FROM pg_inherits
+        JOIN pg_class parent ON pg_inherits.inhparent = parent.oid
+        JOIN pg_class child  ON pg_inherits.inhrelid  = child.oid
+        JOIN pg_namespace    ON parent.relnamespace = pg_namespace.oid
+      WHERE parent.relname = #{quote(table)}
+        AND pg_namespace.nspname = #{quote(schema)}
+    SQL
```
rkrage, line 203: Should we validate that child tables themselves are _not_ partitioned tables? Currently, if we are in a sub-partitioning scenario, the call to `safe_add_concurrent_index` below would raise an error and we'd be left in a state with an invalid index on the parent table

lib/pg_ha_migrations/safe_statements.rb
jcoleman: Yes, I think we should support that (Rails does).

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -170,6 +170,65 @@ def safe_remove_concurrent_index(table, options={})
...
+
+    parent_index = "#{table}_#{name_suffix}"
+
+    quoted_columns = Array.wrap(columns).map { |c| quote_column_name(c) }.join(", ")
```
jcoleman, line 207: Does this exclude being able to create function/expression indexes.

lib/pg_ha_migrations/safe_statements.rb
jcoleman: Yes, I think so.

Actually, that could be checked in `safe_add_concurrent_index` itself, which would be even more broadly useful.

I also think it might be worth exposing this block as a public method (`def children_of_partition_parent_table(table)`?); I've definitely need to use something similar before.

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -170,6 +170,65 @@ def safe_remove_concurrent_index(table, options={})
...
+      SELECT child.relname
+      FROM pg_inherits
+        JOIN pg_class parent ON pg_inherits.inhparent = parent.oid
+        JOIN pg_class child  ON pg_inherits.inhrelid  = child.oid
+        JOIN pg_namespace    ON parent.relnamespace = pg_namespace.oid
+      WHERE parent.relname = #{quote(table)}
+        AND pg_namespace.nspname = #{quote(schema)}
+    SQL
+
+    parent_index = "#{table}_#{name_suffix}"
```
jcoleman, line 205: Do we need to handle long names like ActiveRecord does?

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -170,6 +170,65 @@ def safe_remove_concurrent_index(table, options={})
...
+
+    unsafe_execute(<<~SQL)
+      CREATE INDEX#{if_not_exists ? " IF NOT EXISTS" : ""} #{quote_column_name(parent_index)}
+      ON ONLY #{quote_schema_name(schema)}.#{quote_table_name(table)} (#{quoted_columns})
+    SQL
+
+    child_tables.each do |child_table|
+      child_index = "#{child_table}_#{name_suffix}"
+
+      next if if_not_exists && _index_exists?(schema, child_index)
```
jcoleman, line 217: Can we use "normal" conditional style? I.e.:

```
unless if_not_exists && _index_exists?(schema, child_index)
  # do the thing
end
```

Also...do we want to use a real Postgres `IF NOT EXISTS`?

README.md
```diff
@@ -188,6 +188,34 @@ Safely remove an index. Migrations that contain this statement must also include
...
+
+```ruby
+# Assuming this table has partitions child1 and child2, the following indexes will be created:
+#   - partitioned_table_index_suffix
+#   - child1_index_suffix (attached to partitioned_table_index_suffix)
+#   - child2_index_suffix (attached to partitioned_table_index_suffix)
+safe_add_concurrent_partitioned_index :partitioned_table, [:column1, :column2], name_suffix: "index_suffix"
+```
+
+Note: this method does not support sub-partitioning.
```
jcoleman, line 217: Should we document that this method isn't guaranteed to complete all the work? I.e., it can fail after successfully completing some children but not all?

It's also not transactional per index, but we probably need to document that on `safe_add_concurrent_index` itself.

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -170,6 +170,65 @@ def safe_remove_concurrent_index(table, options={})
...
+    SQL
+
+    parent_index = "#{table}_#{name_suffix}"
+
+    quoted_columns = Array.wrap(columns).map { |c| quote_column_name(c) }.join(", ")
+
+    unsafe_execute(<<~SQL)
+      CREATE INDEX#{if_not_exists ? " IF NOT EXISTS" : ""} #{quote_column_name(parent_index)}
+      ON ONLY #{quote_schema_name(schema)}.#{quote_table_name(table)} (#{quoted_columns})
+    SQL
```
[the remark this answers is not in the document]
jcoleman: I think probably this one needs a `say_with_time`, but the other ones we should just allow the output from `safe_add_concurrent_index` to come through.

lib/pg_ha_migrations/safe_statements.rb
jcoleman: We need to wrap it in `safe_acquire_lock_for_table` (and a `say_with_time` I think).

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -170,6 +170,65 @@ def safe_remove_concurrent_index(table, options={})
...
+      safe_add_concurrent_index("#{quote_schema_name(schema)}.#{child_table}", columns, name: child_index)
+
+      unsafe_execute(<<~SQL)
+        ALTER INDEX #{quote_schema_name(schema)}.#{quote_column_name(parent_index)}
+        ATTACH PARTITION #{quote_schema_name(schema)}.#{quote_column_name(child_index)}
+      SQL
+    end
+
+    if !_index_exists?(schema, parent_index, check_validity: true)
+      raise PgHaMigrations::InvalidMigrationError, "Parent index #{parent_index} is invalid"
```
jcoleman, line 228: This should never happen, right?

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -402,37 +461,51 @@ def _quoted_partman_schema
   end
 
   def _fully_qualified_table_name_for_partman(table)
+    _schema_and_table_for(table).each do |identifier|
+      if identifier.to_s !~ /^[a-z_][a-z_\d]*$/
+        raise PgHaMigrations::InvalidMigrationError, "Partman requires schema / table names to be lowercase with underscores"
```
jcoleman, line 466: Could this include the invalid name in its output? And if we do that, we probably need to not short-circuit (i.e., we want to know if _both_ parts are invalid).

spec/safe_statements_spec.rb
```diff
@@ -1263,6 +1281,239 @@ def up
...
+              def up
+                safe_add_concurrent_partitioned_index :foos3, :updated_at
+              end
+            end
+
+            allow(ActiveRecord::Base.connection).to receive(:execute).and_call_original
+
+            test_migration.suppress_messages { test_migration.migrate(:up) }
+
+            expect(ActiveRecord::Base.connection).to have_received(:execute).with(/CREATE INDEX/).once
```
jcoleman, line 1305: Personal preference: could we use the `to receive` style (not the `to have_received` style)?

spec/safe_statements_spec.rb
```diff
@@ -1263,6 +1281,239 @@ def up
...
+
+            allow(ActiveRecord::Base.connection).to receive(:execute).and_call_original
+
+            test_migration.suppress_messages { test_migration.migrate(:up) }
+
+            expect(ActiveRecord::Base.connection).to have_received(:execute).with(/CREATE INDEX/).exactly(11).times
+            expect(ActiveRecord::Base.connection).to have_received(:execute).with(/ALTER INDEX/).exactly(10).times
+
+            # look up indexes on child tables and parent table
+            partitions_for_table(:foos3).append(:foos3).each do |table|
```
jcoleman, line 1335: Can we wrap this in `aggregate_failures`?

spec/safe_statements_spec.rb
```diff
@@ -1263,6 +1281,239 @@ def up
...
+              def up
+                safe_add_concurrent_partitioned_index :foos3, :updated_at, if_not_exists: true
+              end
+            end
+
+            allow(ActiveRecord::Base.connection).to receive(:execute).and_call_original
+
+            test_migration.suppress_messages { test_migration.migrate(:up) }
+
+            expect(ActiveRecord::Base.connection).to have_received(:execute).with(/CREATE INDEX/).once
```
jcoleman, line 1368: Can we wrap this in `aggregate_failures`?

spec/safe_statements_spec.rb
```diff
@@ -1263,6 +1281,239 @@ def up
...
+
+            allow(ActiveRecord::Base.connection).to receive(:execute).and_call_original
+
+            test_migration.suppress_messages { test_migration.migrate(:up) }
+
+            expect(ActiveRecord::Base.connection).to have_received(:execute).with(/CREATE INDEX/).twice
+            expect(ActiveRecord::Base.connection).to have_received(:execute).with(/ALTER INDEX/).once
+
+            # look up indexes on child tables and parent table
+            ["foos3'", "foos3'_child"].each do |table|
```
jcoleman, line 1403: Can we wrap this in `aggregate_failures`?

lib/pg_ha_migrations/safe_statements.rb
rkrage: So the `CREATE INDEX ... ON ONLY` happens first before we add the indexes on child tables, so if we add the validation to `safe_add_concurrent_index` we'd still be left with an invalid index on the parent table. I think maybe it makes to have that validation in both this method and `safe_add_concurrent_index`. I'll start working on that

> I also think it might be worth exposing this block as a public method (def children_of_partition_parent_table(table)?

Yeah that makes sense, but I think I'll open a separate PR

lib/pg_ha_migrations/safe_statements.rb
rkrage: Oh I didn't realize there was a character limit, so probably, yes. I'll try to look at the rails code to see what they're doing.

lib/pg_ha_migrations/safe_statements.rb
rkrage: Yeah we're assuming `columns` is either a single column or array of columns. Do we want to support expressions?

lib/pg_ha_migrations/safe_statements.rb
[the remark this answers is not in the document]
rkrage: In addition to the `safe_add_concurrent_index` calls, we also have calls to `ALTER INDEX ... ATTACH PARTITION` though

lib/pg_ha_migrations/safe_statements.rb
rkrage: I prefer to only use `unless` for single line conditionals. Otherwise it reads strangely in my head. But sure, I can try to rework this logic to avoid using `next`

I _think_ the `ALTER INDEX ... ATTACH PARTITION` call will fail if the index already exists and has been attached, which is why I chose to explicitly check for existence up front. I'll try to dig into that a bit more

lib/pg_ha_migrations/safe_statements.rb
rkrage: With doing no research whatsoever, my gut tells me that the implementation for partitioned indexes in postgres is fairly complicated and there are probably some conditions we're not thinking about that would cause the index to be invalid. In theory, this should never happen, but I think it's best to err on the side of caution and double check that everything worked as expected. We absolutely do not want a silent failure where this method succeeds, the index is invalid, and the user has no idea.

lib/pg_ha_migrations/safe_statements.rb
rkrage: Yeah I can look into that

spec/safe_statements_spec.rb
rkrage: So I _tried_ using `to receive` but I think rspec gets confused when you have multiple expectations on the same method with different arguments.

spec/safe_statements_spec.rb
rkrage: TIL!

lib/pg_ha_migrations/safe_statements.rb
rkrage: So apparently whatever website I was looking at was wrong (I can't even find it at this point). After further investigation, on PG 11-13 `ALTER INDEX ... ATTACH PARTITION` takes out an `AccessShareLock` on the child table and on PG 14-15 it takes out `AccessShareLock` on the child _and_ parent table (or at least that's what is being displayed in the `pg_locks` view).

Considering issue https://github.com/braintree/pg_ha_migrations/issues/39 hasn't been implemented yet, I don't think we want `safe_acquire_lock_for_table` here

lib/pg_ha_migrations/safe_statements.rb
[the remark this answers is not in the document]
rkrage: For those curious, this takes out a `ShareLock` on the parent table, just like any other `CREATE INDEX` operation (without `CONCURRENTLY`). This should be crazy fast though since it's not actually building the index, so I _don't think_ we need a call to `safe_acquire_lock_for_table` here

lib/pg_ha_migrations/safe_statements.rb
rkrage: After digging through rails code, there's actually quite a bit of logic needed to support expressions. Since the native `add_index` method doesn't support `ON ONLY` and we have to build that query ourselves, we'd have to duplicate that expression logic here. So that this PR doesn't balloon out of control, I think it's okay to implement a subset of the features `add_index` supports and enhance it in the future if needed

lib/pg_ha_migrations/safe_statements.rb
rkrage: It turns out `ALTER INDEX ... ATTACH PARTITION` does _not_ fail if the index is already attached, so we _could_ use the native `IF NOT EXISTS`. It would just result in some extra queries. Do you think that's a reasonable tradeoff to simplify the code?

rkrage: Soooo I tried making this change but it appears there's a bug in rails 6.1 when generating the SQL (`CONCURRENTLY` should be before `IF NOT EXISTS`):

```
PG::SyntaxError:
  ERROR:  syntax error at or near "CONCURRENTLY"
  LINE 1: CREATE INDEX IF NOT EXISTS CONCURRENTLY "foos3_p2023w29_upda...
```

lib/pg_ha_migrations/safe_statements.rb
jcoleman: > What should the message be for say_with_time though?

I realize we don't have a method name it can output, but it seems like it's still an operation we should expose (it is DDL after all); maybe something like `Attaching per-partition index #{index name} to parent #{index name}`?

jcoleman: > After further investigation, on PG 11-13 ALTER INDEX ... ATTACH PARTITION takes out an AccessShareLock on the child table and on PG 14-15 it takes out AccessShareLock on the child and parent table (or at least that's what is being displayed in the pg_locks view).

We need to do more verification (and perhaps code spelunking). According to the [Postgres docs](https://www.postgresql.org/docs/current/sql-alterindex.html):

> the lock level required may differ for each subform. An ACCESS EXCLUSIVE lock is held unless explicitly noted.

and the `ATTACH PARTITION` subform has no such note. That being said it would very much surprise me to find that it's doing a lock escalation, since that usually is a vector for deadlocks.

lib/pg_ha_migrations/safe_statements.rb
jcoleman: Yeah, that's fine, but I think we should make that clear in the message then. E.g., something like "Unexpected state: <what you already have>"

lib/pg_ha_migrations/safe_statements.rb
jcoleman: If you're keeping track of that, that's good for me.

lib/pg_ha_migrations/safe_statements.rb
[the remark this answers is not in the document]
rkrage: We spoke offline, but for context:

This cannot be done with `CONCURRENTLY`, but it feels like maybe it should. We should hop on the mailing list to see if there's a technical reason preventing this, and if not make a contribution to postgres core

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -170,6 +175,87 @@ def safe_remove_concurrent_index(table, options={})
     unsafe_remove_index(table, **options.merge(:algorithm => :concurrently))
   end
 
+  # TODO: figure out what options we actually want to allow
+  #
+  # we definitely don't want algorithm and unique
+  def safe_add_concurrent_partitioned_index(table, columns, name_suffix:, if_not_exists: false, using: nil)
+    raise ArgumentError, "Expected <name_suffix> to be present" unless name_suffix.present?
```
rkrage, line 182: We need to lookup the table schema to accurately run discovery queries (looking up child partitions / checking index validity), and therefore always use the fully qualified table name when calling `add_index`. Rails will generate a name like `public.foo_column_idx` if the index name isn't explicitly provided, which isn't expected if you call this method as `safe_add_concurrent_partitioned_index :foo, :column`. Additionally, we need the name of the child indexes when calling `ALTER INDEX ... ATTACH PARTITION` (`add_index` returns a `PG::Result` object so we probably _could_ get the index name dynamically, but API doesn't explicitly call out the return type for this method so I don't think we should rely on that).

With all that being said, to simplify this code I think it's perfectly reasonable to just require `name_suffix` to be present

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -154,6 +155,10 @@ def unsafe_make_column_not_nullable(table, column, options={}) # options arg is
     end
   end
 
+  # In early versions of Rails 6.0 (fixed in 6.1.4), there is a bug in the SQL
+  # generation when if_not exists is present: https://github.com/rails/rails/pull/41490
+  #
+  # Should we backport a fix in our hacks, raise an error, or do nothing?
```
rkrage, line 161: Left a code comment here just as a reminder to myself to add a comment on the PR and start a discussion 😄

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -170,6 +175,87 @@ def safe_remove_concurrent_index(table, options={})
     unsafe_remove_index(table, **options.merge(:algorithm => :concurrently))
   end
 
+  # TODO: figure out what options we actually want to allow
+  #
+  # we definitely don't want algorithm and unique
```
rkrage, line 180: Another comment to start a PR discussion

lib/pg_ha_migrations/safe_statements.rb
rkrage: This is irrelevant now. The latest commit monkey patches `add_index` to allow `ON ONLY`

lib/pg_ha_migrations/safe_statements.rb
rkrage: We talked about this one offline as well:

Not sure why I wasn't able to observe the `AccessExclusiveLock` in testing, but the source code makes it pretty clear what's happening: https://github.com/postgres/postgres/blob/c0dc97c7bfd14b460536c40d14787f82ab61d0d9/src/backend/commands/tablecmds.c#L18981

The latest commit wraps all of the `ATTACH PARTITION` calls in a `safely_acquire_lock_for_table` block

lib/pg_ha_migrations/safe_statements.rb
rkrage: I think my opinion is to do nothing ™️ 

I don't think it's actively _harmful_, so if a user runs into this bug, they should just bump the patch version of rails, which is generally pretty straightforward. Maybe just add a note in the readme?

lib/pg_ha_migrations/safe_statements.rb
rkrage: I addressed this in [f882261](https://github.com/braintree/pg_ha_migrations/pull/85/commits/f882261491c0a8a72490fdc44ea3d75daf958efa) which allows you to specify `unique`, `where`, and `comment`. 

A few notes:
- A comment will only be added to the parent index (seems redundant to add it to the child indexes)
- You can create a unique index if it includes the partition key column(s), but it's hard to validate this before executing the DDL, so using this option may result in a PG error that bubbles up like so:

  > PG::FeatureNotSupported:
  ERROR:  unique constraint on partitioned table must include all partitioning columns
  DETAIL:  UNIQUE constraint on table "foos3" lacks column "created_at" which is part of the partition key.

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -170,6 +175,99 @@ def safe_remove_concurrent_index(table, options={})
...
+
+      child_index = "#{child_table}_#{name_suffix}"
+
+      _validate_index_name!(child_index)
+
+      [child_schema, child_table, child_index]
+    end
+
+    # TODO: take out ShareLock after issue #39 is implemented
+    safely_acquire_lock_for_table(fully_qualified_parent_table) do
```
rkrage, line 225: If there are blocking queries running against table `foo` in the `public` schema, and you call this method as `safely_acquire_lock_for_table("public.foo") { ... }`, it won't actually match anything here since `tables_with_locks` doesn't contain schema-qualified table names: https://github.com/braintree/pg_ha_migrations/blob/d7ad2e802dae0ffdda62863b495469dd7ec1517d/lib/pg_ha_migrations/safe_statements.rb#L476

Additionally, `safely_acquire_lock_for_table` will not check for locks against partitioned tables since we're only looking at `relkind = 'r'`: https://github.com/braintree/pg_ha_migrations/blob/d7ad2e802dae0ffdda62863b495469dd7ec1517d/lib/pg_ha_migrations/blocking_database_transactions.rb#L59

It _might_ be okay to merge this PR without addressing this issue, but we 100% need to solve this before releasing a new version of the gem (I'm currently working on a fix)

lib/pg_ha_migrations/safe_statements.rb
jcoleman: Yeah, adding a note to the readme would be good. If we wanted to we could catch that specific error and re-raise a message about that, but it's probably not worth it.

spec/safe_statements_spec.rb
rkrage: Fixed in https://github.com/braintree/pg_ha_migrations/pull/85/commits/a473f1fb102e8882d1ba0fbc11cb038e5e3bcb5a

lib/pg_ha_migrations/safe_statements.rb
rkrage: Added some notes in https://github.com/braintree/pg_ha_migrations/pull/85/commits/dba97ce4350453a84b970f7b0855d2f5f3615f9f

spec/safe_statements_spec.rb
```diff
@@ -1263,6 +1281,552 @@ def up
...
+            end
+
+            allow(ActiveRecord::Base.connection).to receive(:execute).and_call_original
+
+            test_migration.suppress_messages { test_migration.migrate(:up) }
+
+            aggregate_failures do
+              expect(ActiveRecord::Base.connection).to have_received(:execute).with(/CREATE INDEX "foos3'_bar""" ON ONLY/).once
+              expect(ActiveRecord::Base.connection).to have_received(:execute).with(/CREATE INDEX CONCURRENTLY "foos3'_child_bar""" ON/).once
+              expect(ActiveRecord::Base.connection).to have_received(:execute).with(/ALTER INDEX/).once
```
jcoleman: This should probably match `ATTACH` also.

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -170,6 +175,99 @@ def safe_remove_concurrent_index(table, options={})
...
+      # CREATE INDEX ON ONLY parent_table
```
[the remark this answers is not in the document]
rkrage: I think the main goal of this method is to be able to concurrently create an index when data already exists in a partitioned table. I could maybe see adding a warning message that says, "hey this is gonna take awhile", if there's a ton of data. But we don't really have a precedent for that, as there are plenty of other operations in this library that could take a really long time

For this particular call within this method, I _don't think_ the size of the data really impacts the speed in which `CREATE INDEX ON ONLY` executes

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -170,6 +175,99 @@ def safe_remove_concurrent_index(table, options={})
...
+        if_not_exists: if_not_exists,
+        using: using,
+        unique: unique,
+        where: where,
+      )
+    end
+
+    # Avoid taking out an unnecessary lock when there are
+    # no child tables or the index is already valid
+    if child_tables_with_metadata.present? && !_index_valid?(parent_schema, parent_index)
```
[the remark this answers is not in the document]
rkrage: This could happen if the initial method call falls for whatever reason and you re-execute with `if_not_exists: true`

lib/pg_ha_migrations/hacks/add_index_on_only.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+
+    module CreateIndexDefinition
+      def visit_CreateIndexDefinition(o)
+        if o.algorithm == "ONLY"
+          o.algorithm = nil
+
+          original_query = super
+
+          if original_query.gsub(" ON ").count > 1
```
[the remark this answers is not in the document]
rkrage: The original idea was to prevent a really obscure situation where you have a table or column name like `"foo ON bar"` (and there's a test for that), but agree it could maybe be too restrictive.  We could maybe replace `ON <table name>` with `ON ONLY <table name>`?

README.md
```diff
@@ -188,6 +188,38 @@ Safely remove an index. Migrations that contain this statement must also include
...
+#   - child1_column1_column2_idx (attached to partitioned_table_column1_column2_idx)
+#   - child2_column1_column2_idx (attached to partitioned_table_column1_column2_idx)
+safe_add_concurrent_partitioned_index :partitioned_table, [:column1, :column2], name_suffix: "column1_column2_idx", using: :hash
+```
+
+Notes:
+
+- this method does not support sub-partitioning
+- this method runs multiple DDL statements non-transactionally
+  - creating / attaching an index on a child table could fail, resulting in an error and an invalid index on the parent table
```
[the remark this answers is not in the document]
rkrage: > In the case such an exception is raised an `INVALID` index will be left on the parent table.

That reads pretty strangely to me

lib/pg_ha_migrations/hacks/add_index_on_only.rb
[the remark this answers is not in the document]
rkrage: This seems a bit better to me: https://github.com/braintree/pg_ha_migrations/pull/85/commits/22e74dab8ff40965bbe4be515122c4419dc7ab91

spec/safe_statements_spec.rb
```diff
@@ -34,12 +36,17 @@ def grantees_for_table(table)
     SQL
   end
 
-  def create_range_partitioned_table(table, migration_klass, with_template: false)
+  def create_range_partitioned_table(table, migration_klass, with_template: false, with_partman: false)
     migration = Class.new(migration_klass) do
-      class_attribute :table, :with_template, instance_accessor: true
+      class_attribute \
```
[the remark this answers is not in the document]
rkrage: I just brought it back to a single line in b22c05d

spec/safe_statements_spec.rb
rkrage: Done in b22c05d

README.md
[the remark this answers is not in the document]
jcoleman: Tweaked the suggestion.

lib/pg_ha_migrations/hacks/add_index_on_only.rb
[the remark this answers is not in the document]
jcoleman: Do we know that this has to start with `CREATE INDEX[ CONCURRENTLY] `? Could we anchor to the front of string with that?

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -170,6 +175,99 @@ def safe_remove_concurrent_index(table, options={})
...
+      FROM pg_inherits
+        JOIN pg_class parent        ON pg_inherits.inhparent = parent.oid
+        JOIN pg_class child         ON pg_inherits.inhrelid  = child.oid
+        JOIN pg_namespace parent_ns ON parent.relnamespace = parent_ns.oid
+        JOIN pg_namespace child_ns  ON child.relnamespace = child_ns.oid
+      WHERE parent.relname = #{connection.quote(parent_table)}
+        AND parent_ns.nspname = #{connection.quote(parent_schema)}
+    SQL
+
+    child_tables_with_metadata = child_schemas_and_tables.map do |child_schema, child_table|
```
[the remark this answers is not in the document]
jcoleman: Heh, my reasoning was mostly to make it more obvious in the code internally.

lib/pg_ha_migrations/safe_statements.rb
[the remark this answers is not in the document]
jcoleman: Oh wait, are we guaranteeing this is a natively partitioned table already (and not a partman partitioned one)? If so, I think we'd be fine, because the "partition" table _when natively partitioned_ is [guaranteed to be empty](https://www.postgresql.org/docs/current/ddl-partitioning.html):

> The partitioned table itself is a “virtual” table having no storage of its own.

But if we support partman partitioned tables than the partition parent itself can have data, and we're not running this `concurrently`, so it could be a long-held lock.

lib/pg_ha_migrations/safe_statements.rb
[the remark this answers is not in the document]
jcoleman: Do we have a test for that? Even if so, it seems like that would be better handled by checking to see if a valid index exists on the partition parent up-front rather than here at the end.

lib/pg_ha_migrations/safe_statements.rb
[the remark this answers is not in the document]
jcoleman: TBC: I won't thumbs down based on this; I just think it'd likely make it easier to follow the code in the future.

lib/pg_ha_migrations/hacks/add_index_on_only.rb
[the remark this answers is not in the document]
rkrage: You would also need to account for `UNIQUE` as well. I think what I have currently is pretty robust. It would be incredibly rare for `<index name> ON <table name>` to show up more than once in the query string. And even if it does, it would probably be in a where clause somewhere. Since we're doing `sub` and not `gsub` it would correctly replace the first occurrence at the start of the string

lib/pg_ha_migrations/safe_statements.rb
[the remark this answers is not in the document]
rkrage: Yeah, we're querying `pg_partitioned_table` so this method will only target natively partitioned tables. I actually don't know if attaching an index is something you can do with normal table inheritance... But either way, I think it's fine to just target natively partitioned tables in this iteration (the readme already calls this out)

lib/pg_ha_migrations/safe_statements.rb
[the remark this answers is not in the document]
rkrage: That seems reasonable, but I know there was a reason why I did this... Lemme think about it and get back to you

lib/pg_ha_migrations/safe_statements.rb
rkrage: Addressed in https://github.com/braintree/pg_ha_migrations/pull/85/commits/71c773df9e90014afd2c4ae82c3c253ffa04fcc1

lib/pg_ha_migrations/safe_statements.rb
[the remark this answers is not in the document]
rkrage: Yeah... I can't think of a reason why we wouldn't just do this check up-front. I thought that maybe there would be a situation where an index could exist with the same name (with different columns / uniqueness) and therefore we would want to let the `CREATE INDEX` calls fail. However, those calls will _not_ fail. It appears `IF NOT EXISTS` simply checks for the existence of an index matching the identifier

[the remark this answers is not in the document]
rkrage: Should be addressed in https://github.com/braintree/pg_ha_migrations/pull/85/commits/31cc86bd6194896a5cd88c77eea5b70586ed9132

lib/pg_ha_migrations/safe_statements.rb
jcoleman: Do we want name suffix? Or just name, use name for the index on the parent, and do whatever we want for the child tables?

One other thing: I think it's OK to require the name to match for the `if_not_exists` (it saves a lot of effort on trying to lookup indexes and match them opportunistically), but I wonder if we should explicitly document as such.

lib/pg_ha_migrations/hacks/add_index_on_only.rb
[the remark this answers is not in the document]
jcoleman: Oh, good point on "first occurrence".

lib/pg_ha_migrations/safe_statements.rb
rkrage: > Do we want name suffix? Or just name, use name for the index on the parent, and do whatever we want for the child tables?

That's a really good point because Postgres will generate its own index name for child partitions that get added in the future, so it will be pretty difficult to ensure consistent child index names when doing time-based partitioning. I'll push a commit shortly

> One other thing: I think it's OK to require the name to match for the if_not_exists (it saves a lot of effort on trying to lookup indexes and match them opportunistically), but I wonder if we should explicitly document as such.

I believe the base `add_index` Rails method also behaves this way, so since it's consistent I don't think we need explicit documentation

jcoleman (approved):
Let's get this squashed down into meaningful commits, but otherwise LGTM.

Thanks for working through the complexity of this.

PR #86, opened 2023-09-05
Support to recursively create partitioned index

rkrage:
Recurse through sub-partitions when creating a partitioned index

Relies on https://github.com/braintree/pg_ha_migrations/pull/94

### DIFF
--- a/lib/pg_ha_migrations/safe_statements.rb
+++ b/lib/pg_ha_migrations/safe_statements.rb
@@ -199,8 +199,6 @@ module PgHaMigrations::SafeStatements
     return if if_not_exists && parent_index.valid?
 
     child_indexes = parent_table.partitions.map do |child_table|
-      raise PgHaMigrations::InvalidMigrationError, "Partitioned table #{parent_table.inspect} contains sub-partitions" if child_table.natively_partitioned?
-
       PgHaMigrations::Index.from_table_and_columns(child_table, columns)
     end
 
@@ -221,8 +219,14 @@ module PgHaMigrations::SafeStatements
     end
 
     child_indexes.each do |child_index|
-      # CREATE INDEX CONCURRENTLY ON child_table
-      safe_add_concurrent_index(
+      add_index_method = if child_index.table.natively_partitioned?
+        :safe_add_concurrent_partitioned_index
+      else
+        :safe_add_concurrent_index
+      end
+
+      send(
+        add_index_method,
         child_index.table.fully_qualified_name,
         columns,
         name: child_index.name,
@@ -490,7 +494,6 @@ module PgHaMigrations::SafeStatements
       end
     end.join(".")
   end
-
 
   def _per_migration_caller
     @_per_migration_caller ||= Kernel.caller
--- a/spec/safe_statements_spec.rb
+++ b/spec/safe_statements_spec.rb
@@ -1372,8 +1372,11 @@ RSpec.describe PgHaMigrations::SafeStatements do
             test_migration.suppress_messages { test_migration.migrate(:up) }
 
             aggregate_failures do
-              # look up indexes on child tables and parent table
-              partitions_for_table(:foos3).append(:foos3).each do |table|
+              tables_with_indexes = partitions_for_table(:foos3).append(:foos3)
+
+              expect(tables_with_indexes.size).to eq(11)
+
+              tables_with_indexes.each do |table|
                 indexes = ActiveRecord::Base.connection.indexes(table)
 
                 expected_comment = table == :foos3 ? "this is an index" : nil
@@ -1411,8 +1414,11 @@ RSpec.describe PgHaMigrations::SafeStatements do
             test_migration.suppress_messages { test_migration.migrate(:up) }
 
             aggregate_failures do
-              # look up indexes on child tables and parent table
-              partitions_for_table(:foos3).append(:foos3).each do |table|
+              tables_with_indexes = partitions_for_table(:foos3).append(:foos3)
+
+              expect(tables_with_indexes.size).to eq(11)
+
+              tables_with_indexes.each do |table|
                 indexes = ActiveRecord::Base.connection.indexes(table)
 
                 expect(indexes.size).to eq(1)
@@ -1446,8 +1452,11 @@ RSpec.describe PgHaMigrations::SafeStatements do
             test_migration.suppress_messages { test_migration.migrate(:up) }
 
             aggregate_failures do
-              # look up indexes on child tables and parent table
-              partitions_for_table(:foos3).append(:foos3).each do |table|
+              tables_with_indexes = partitions_for_table(:foos3).append(:foos3)
+
+              expect(tables_with_indexes.size).to eq(11)
+
+              tables_with_indexes.each do |table|
                 indexes = ActiveRecord::Base.connection.indexes(table)
 
                 expect(indexes.size).to eq(1)
@@ -1482,8 +1491,11 @@ RSpec.describe PgHaMigrations::SafeStatements do
             test_migration.suppress_messages { test_migration.migrate(:up) }
 
             aggregate_failures do
-              # look up indexes on child tables and parent table
-              partitions_for_table(:foos3).append(:foos3).each do |table|
+              tables_with_indexes = partitions_for_table(:foos3).append(:foos3)
+
+              expect(tables_with_indexes.size).to eq(11)
+
+              tables_with_indexes.each do |table|
                 indexes = ActiveRecord::Base.connection.indexes(table)
 
                 expect(indexes.size).to eq(1)
@@ -1518,8 +1530,11 @@ RSpec.describe PgHaMigrations::SafeStatements do
             test_migration.suppress_messages { test_migration.migrate(:up) }
 
             aggregate_failures do
-              # look up indexes on child tables and parent table
-              partitions_for_table(:foos3).append(:foos3).each do |table|
+              tables_with_indexes = partitions_for_table(:foos3).append(:foos3)
+
+              expect(tables_with_indexes.size).to eq(11)
+
+              tables_with_indexes.each do |table|
                 indexes = ActiveRecord::Base.connection.indexes(table)
 
                 expect(indexes.size).to eq(1)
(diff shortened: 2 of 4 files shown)


lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -183,7 +183,8 @@ def safe_add_concurrent_partitioned_index(
     using: nil,
     unique: nil,
     where: nil,
-    comment: nil
+    comment: nil,
+    recurse: nil
```
jcoleman, line 187: When would we want `recurse` to be false?

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -200,16 +201,47 @@ def safe_add_concurrent_partitioned_index(
...
+          using: using,
+          unique: unique,
+          where: where,
+        )
+      elsif !if_not_exists
+        message = "Partitioned table #{parent_table.inspect} contains sub-partitions. " \
+          "If sub-partition indexes have already been created, execute this method with 'if_not_exists: true'. " \
+          "If sub-partition indexes have not been created, consider executing this method with 'recurse: true'."
+
+        raise PgHaMigrations::InvalidMigrationError, message
```
jcoleman, line 236: This is taking a hint (hehe) from Postgres's error style, but:
```suggestion
        raise PgHaMigrations::InvalidMigrationError, <<~EOS
          Partitioned table #{parent_table.inspect} contains sub-partitions.
          Hint: If sub-partition indexes have already been created, execute this method with 'if_not_exists: true'.
          Hint: If sub-partition indexes have not been created, consider executing this method with 'recurse: true'.
        EOS
```

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -200,16 +201,47 @@ def safe_add_concurrent_partitioned_index(
...
+      elsif !_index_valid?(child_schema, child_index)
+        message = "Sub-partition #{child_table.inspect} does not have valid index #{child_index.inspect}. " \
+          "Consider executing this method with 'recurse: true'."
+
+        raise PgHaMigrations::InvalidMigrationError, message
```
jcoleman, line 241:
```suggestion
        raise PgHaMigrations::InvalidMigrationError, <<~EOS
          Sub-partition #{child_table.inspect} does not have valid index #{child_index.inspect}.
          Hint: Consider executing this method with 'recurse: true'.
```

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -227,6 +259,11 @@ def safe_add_concurrent_partitioned_index(
     end
 
     child_tables_with_metadata.each do |child_schema, child_table, child_index|
+      # If we get past all of the validation above, we can safely assume that
+      # indexes on sub-partitions have already been created recursviely or in a
+      # separate, explicit call to safe_add_concurrent_partitioned_index
+      next if _partitioned_table?(child_schema, child_table)
```
jcoleman, line 265: Same thing here :D

spec/safe_statements_spec.rb
```diff
@@ -1692,24 +1692,116 @@ def up
...
+                  table: table,
+                  name: "#{table}_updated_at_idx",
+                  columns: ["updated_at"],
+                  using: :btree,
+                )
+              end
+            end
+          end
+
+          it "raises error when sub-partitioning detected and if_not_exists is false" do
```
jcoleman, line 1794: Why does sub-partitioning imply we must use `if_not_exists`?

spec/safe_statements_spec.rb
```diff
@@ -1723,10 +1815,10 @@ def up
 
             expect do
               test_migration.suppress_messages { test_migration.migrate(:up) }
-            end.to raise_error(PgHaMigrations::InvalidMigrationError, "Table \"foos3\" is not a partitioned table")
+            end.to raise_error(PgHaMigrations::InvalidMigrationError, /Partitioned table "foos3" contains sub-partitions/)
           end
 
-          it "raises error when sub-partitioning detected" do
+          it "raises error when sub-partition indexes have not been created and if_not_exists is true" do
```
jcoleman, line 1821: Is this also `recurse` is false?

spec/safe_statements_spec.rb
rkrage: I think the test description is just wrong and it should be:

> raises error when sub-partitioning detected, recurse is false, and if_not_exists is false

spec/safe_statements_spec.rb
rkrage: Yup, I'll update the description

lib/pg_ha_migrations/safe_statements.rb
rkrage: So I'm imagining a scenario where the user might want to break up the operation into several discrete method calls (or even multiple migrations). If you have sub-partitioning that is multiple levels deep, maybe you'd want to do:

```ruby
safe_add_concurrent_partitioned_index :foo_sub_level_2, :column, name_suffix: :column
safe_add_concurrent_partitioned_index :foo_sub_level_1, :column, name_suffix: :column, if_not_exists: true
safe_add_concurrent_partitioned_index :foo, :column, name_suffix: :column, if_not_exists: true
```

Of course, I think you could accomplish the same thing with:

```ruby
safe_add_concurrent_partitioned_index :foo_sub_level_2, :column, name_suffix: :column, recurse: true
safe_add_concurrent_partitioned_index :foo_sub_level_1, :column, name_suffix: :column, if_not_exists: true, recurse: true
safe_add_concurrent_partitioned_index :foo, :column, name_suffix: :column, if_not_exists: true, recurse: true
```

But there will be significantly more SQL queries being executed in the latter.

Maybe it makes sense to just default `recurse` to `true`, as I think that's what you'd want most of the time, and allow the user to opt out?

jcoleman: Yeah, I like that better. I'm not 100% convinced we _need_ the knob, but I don't really know if I'd want it in the future or not, so I think it's quite reasonable to add it. But doing the reasonable default seems like a good thing up-front.

spec/safe_statements_spec.rb
jcoleman: If we're missing sub partition indexes and `recurse = false` then won't we have an error regardless of the status of `if_not_exists`?

rkrage: Yes, it's just that the error messaging is slightly different

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -200,16 +201,47 @@ def safe_add_concurrent_partitioned_index(
...
       _validate_index_name!(child_index)
 
       [child_schema, child_table, child_index]
     end
 
+    # When sub-partitioning detected:
+    #   - recursively create indexes when recurse is true
+    #   - do nothing when if_not_exists is true and index already valid
+    #   - otherwise raise helpful error
+    child_tables_with_metadata.each do |child_schema, child_table, child_index|
```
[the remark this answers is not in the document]
rkrage: I'm not sure I'm following this one. In both scenarios, if `recurse: true` the result will be a valid index on the parent and if `recurse: false` it will raise an error that sub-partitioning was detected

rkrage:
I rewrote this to just always recurse because it makes the code infinitely cleaner. I still think there are some edge cases to consider, though. Like currently, we `CREATE INDEX ... ON ONLY` upfront before recursing through sub-partitions, which I think is safer because it accounts for a situation where child partitions are added while the operation is running, but it also opens us up to a situation where the parent index gets created, the index name on a sub-partition is invalid due to exceeding the 63 byte limit, an error is raised and we're left with an invalid parent index.

I'm still working through these edge cases and writing tests, so this PR can be considered a draft for now.

spec/safe_statements_spec.rb
```diff
@@ -1501,6 +1501,92 @@ def up
...
+              expect(ActiveRecord::Base.connection).to receive(:execute).with(/CREATE INDEX "index_foos3_sub_on_updated_at" ON ONLY/).once.ordered
+              expect(ActiveRecord::Base.connection).to receive(:execute).with(/ALTER INDEX .+\nATTACH PARTITION/).once.ordered
+              expect(ActiveRecord::Base.connection).to_not receive(:execute).with(/CREATE INDEX CONCURRENTLY/)
+            end
+
+            test_migration.suppress_messages { test_migration.migrate(:up) }
+
+            aggregate_failures do
+              %i[foos3 foos3_sub].each do |table|
+                indexes = ActiveRecord::Base.connection.indexes(table)
```
jcoleman, line 1533: Does this only include valid indexes?

spec/safe_statements_spec.rb
```diff
@@ -1501,6 +1501,92 @@ def up
...
+              expect(ActiveRecord::Base.connection).to receive(:execute).with(/CREATE INDEX "index_foos3_sub_on_updated_at" ON ONLY/).once.ordered
+              expect(ActiveRecord::Base.connection).to receive(:execute).with(/CREATE INDEX CONCURRENTLY/).exactly(10).times.ordered
+              expect(ActiveRecord::Base.connection).to receive(:execute).with(/ALTER INDEX "public"."index_foos3_sub_on_updated_at"\nATTACH PARTITION/).exactly(10).times.ordered
+              expect(ActiveRecord::Base.connection).to receive(:execute).with(/ALTER INDEX "public"."index_foos3_on_updated_at"\nATTACH PARTITION/).exactly(11).times.ordered
+            end
+
+            test_migration.suppress_messages { test_migration.migrate(:up) }
+
+            aggregate_failures do
+              partitions_for_table(:foos3).concat(partitions_for_table(:foos3_sub)).append(:foos3).each do |table|
```
jcoleman, line 1576: I think we should pull these out into variables and also assert the size; otherwise the loop can hide a false green test.

jcoleman (commented):
I'm wondering if we want to have a test or two that shows the the invalid indexes left behind in certain error cases.

spec/safe_statements_spec.rb
rkrage: Yeah that seems reasonable

spec/safe_statements_spec.rb
rkrage: No it would include invalid indexes as well. I think I see what you're getting at, but the subject under test would raise an error if the parent index is invalid, so these indexes are guaranteed to be valid

spec/safe_statements_spec.rb
```diff
@@ -1756,60 +1889,99 @@ def up
...
+                  using: :btree,
+                )
+              end
+
+              expect(ActiveRecord::Base.connection.indexes(:foos3_child2)).to be_empty
+            end
           end
 
-          it "raises error when parent index name is too large" do
+          it "raises error and leaves behind invalid index when child sub-partition index creation missed" do
```
rkrage, line 1936: > I'm wondering if we want to have a test or two that shows the the invalid indexes left behind in certain error cases.

This test and the one above _sort of_ addresses the concern. I think the unexpected state error combined with expectations on the existence / absence of various indexes is enough to confirm the behavior. We could also check the validity of each index that is left behind, but that would involve either using `PgHaMigrations::Index#valid?` (which is something that's already being exercised in this test via the unexpected state error) or introduce a test helper that duplicates the SQL the method executes. I don't see a ton of value there

spec/safe_statements_spec.rb
jcoleman: > the subject under test would raise an error if the parent index is invalid
Sure, but this is the test, and we can't assume the subject 😄

spec/safe_statements_spec.rb
rkrage: I think this was addressed by adding expectations like:

```ruby
expect(ActiveRecord::Base.pluck_from_sql("SELECT indexrelid::regclass FROM pg_index WHERE NOT indisvalid")).to contain_exactly(
  "index_foos3_on_updated_at",
  "index_foos3_sub_on_updated_at",
)
```

jcoleman (approved):

PR #87, opened 2023-09-06
Make safely_acquire_lock_for_table aware of schemas / partitions

rkrage:
As described [here](https://github.com/braintree/pg_ha_migrations/pull/85#discussion_r1317471168), there are a few issues with the locking implementation if a fully qualified table name is provided and/or the table is partitioned. This aims to fix those things and additionally adds some guards to detect nested locks (this is generally bad and can lead to deadlocks).

Also included is a refactor to use structs whenever possible to avoid complicated nested array structures

### DIFF
--- a/lib/pg_ha_migrations/safe_statements.rb
+++ b/lib/pg_ha_migrations/safe_statements.rb
@@ -185,40 +185,32 @@ module PgHaMigrations::SafeStatements
       raise PgHaMigrations::InvalidMigrationError, "Concurrent partitioned index creation not supported on Postgres databases before version 11"
     end
 
-    parent_schema, parent_table = _schema_and_table_for(table)
+    parent_table = PgHaMigrations::Table.from_table_name(table)
 
-    raise PgHaMigrations::InvalidMigrationError, "Table #{parent_table.inspect} is not a partitioned table" unless _partitioned_table?(parent_schema, parent_table)
-
-    fully_qualified_parent_table = "#{connection.quote_schema_name(parent_schema)}.#{parent_table}"
+    raise PgHaMigrations::InvalidMigrationError, "Table #{parent_table.inspect} is not a partitioned table" unless parent_table.natively_partitioned?
 
     parent_index = if name.present?
-      name
+      PgHaMigrations::Index.new(name, parent_table)
     else
-      connection.index_name(parent_table, columns)
+      PgHaMigrations::Index.from_table_and_columns(parent_table, columns)
     end
 
-    connection.send(:validate_index_length!, parent_table, parent_index)
-
     # Short-circuit when if_not_exists: true and index already valid
-    return if if_not_exists && _index_valid?(parent_schema, parent_index)
+    return if if_not_exists && parent_index.valid?
 
-    child_tables_with_metadata = _partitions_for_table(parent_schema, parent_table).map do |child_schema, child_table|
-      raise PgHaMigrations::InvalidMigrationError, "Partitioned table #{parent_table.inspect} contains sub-partitions" if _partitioned_table?(child_schema, child_table)
+    child_indexes = parent_table.partitions.map do |child_table|
+      raise PgHaMigrations::InvalidMigrationError, "Partitioned table #{parent_table.inspect} contains sub-partitions" if child_table.natively_partitioned?
 
-      child_index = connection.index_name(child_table, columns)
-
-      connection.send(:validate_index_length!, child_table, child_index)
-
-      [child_schema, child_table, child_index]
+      PgHaMigrations::Index.from_table_and_columns(child_table, columns)
     end
 
     # TODO: take out ShareLock after issue #39 is implemented
-    safely_acquire_lock_for_table(fully_qualified_parent_table) do
+    safely_acquire_lock_for_table(parent_table.fully_qualified_name) do
       # CREATE INDEX ON ONLY parent_table
       unsafe_add_index(
-        fully_qualified_parent_table,
+        parent_table.fully_qualified_name,
         columns,
-        name: parent_index,
+        name: parent_index.name,
         if_not_exists: if_not_exists,
         using: using,
         unique: unique,
@@ -228,12 +220,12 @@ module PgHaMigrations::SafeStatements
       )
     end
 
-    child_tables_with_metadata.each do |child_schema, child_table, child_index|
+    child_indexes.each do |child_index|
       # CREATE INDEX CONCURRENTLY ON child_table
       safe_add_concurrent_index(
-        "#{connection.quote_schema_name(child_schema)}.#{child_table}",
+        child_index.table.fully_qualified_name,
         columns,
-        name: child_index,
+        name: child_index.name,
         if_not_exists: if_not_exists,
         using: using,
         unique: unique,
@@ -242,20 +234,20 @@ module PgHaMigrations::SafeStatements
     end
 
     # Avoid taking out an unnecessary lock if there are no child tables to attach
-    if child_tables_with_metadata.present?
-      safely_acquire_lock_for_table(fully_qualified_parent_table) do
-        child_tables_with_metadata.each do |child_schema, _, child_index|
+    if child_indexes.present?
+      safely_acquire_lock_for_table(parent_table.fully_qualified_name) do
+        child_indexes.each do |child_index|
           say_with_time "Attaching index #{child_index.inspect} to #{parent_index.inspect}" do
             connection.execute(<<~SQL)
-              ALTER INDEX #{connection.quote_schema_name(parent_schema)}.#{connection.quote_column_name(parent_index)}
-              ATTACH PARTITION #{connection.quote_schema_name(child_schema)}.#{connection.quote_column_name(child_index)}
+              ALTER INDEX #{parent_index.fully_qualified_name}
+              ATTACH PARTITION #{child_index.fully_qualified_name}
             SQL
           end
         end
       end
     end
 
-    raise PgHaMigrations::InvalidMigrationError, "Unexpected state. Parent index #{parent_index.inspect} is invalid" unless _index_valid?(parent_schema, parent_index)
+    raise PgHaMigrations::InvalidMigrationError, "Unexpected state. Parent index #{parent_index.inspect} is invalid" unless parent_index.valid?
   end
 
   def safe_set_maintenance_work_mem_gb(gigabytes)
@@ -439,7 +431,7 @@ module PgHaMigrations::SafeStatements
       retention_keep_table: retention_keep_table,
     }.compact
 
-    unsafe_partman_update_config(create_parent_options[:parent_table], **update_config_options)
+    unsafe_partman_update_config(table, **update_config_options)
   end
 
   def safe_partman_update_config(table, **options)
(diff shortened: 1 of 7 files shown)


spec/safe_statements_spec.rb
```diff
@@ -3484,6 +3496,95 @@ def up
...
             end
 
+            it "does not wait to acquire a lock if same table in different schema is blocked" do
+              stub_const("PgHaMigrations::LOCK_TIMEOUT_SECONDS", 1)
+
+              ActiveRecord::Base.connection.execute("CREATE TABLE partman.bogus_table(pk SERIAL, i INTEGER)")
+
+              begin
+                thread = Thread.new do
+                  migration.safely_acquire_lock_for_table("partman.bogus_table") { sleep 2 }
```
jcoleman, line 3506: I wonder if we should do an explicit lock here rather than relying on our code (since our own code is also what's under test).

spec/safe_statements_spec.rb
```diff
@@ -3484,6 +3496,95 @@ def up
...
+                sleep 1.1
+
+                expect(PgHaMigrations::BlockingDatabaseTransactions).to receive(:find_blocking_transactions)
+                  .at_least(3)
+                  .times
+                  .and_call_original
+
+                migration.suppress_messages do
+                  migration.safely_acquire_lock_for_table(table_name) do
+                    expect(locks_for_table(table_name, connection: alternate_connection)).not_to be_empty
```
jcoleman, line 3546: Should we also expect that there are no explicit locks (not sure if we distinguish) on `bogus_table_default`? If we don't have a way to distinguish I suppose we could expect that `find_blocking_transactions` returns empty before this happens?

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -553,6 +542,31 @@ def _partitioned_table?(schema, table)
...
+    SQL
+
+    if include_sub_partitions
+      sub_partitions = schemas_and_tables.each_with_object([]) do |(child_schema, child_table), arr|
+        if _partitioned_table?(child_schema, child_table)
+          arr.concat(_partitions_for_table(child_schema, child_table, include_sub_partitions: true))
+        end
+      end
+
+      schemas_and_tables.concat(sub_partitions)
```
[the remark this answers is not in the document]
rkrage: Addressed in https://github.com/braintree/pg_ha_migrations/pull/87/commits/bd2c1e92e2cfd7765f579f6baec38bf66f4a1a22

spec/safe_statements_spec.rb
rkrage: Addressed in https://github.com/braintree/pg_ha_migrations/pull/87/commits/bd2c1e92e2cfd7765f579f6baec38bf66f4a1a22

spec/safe_statements_spec.rb
rkrage: Addressed in https://github.com/braintree/pg_ha_migrations/pull/87/commits/bd2c1e92e2cfd7765f579f6baec38bf66f4a1a22

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -587,16 +601,38 @@ def exec_migration(conn, direction)
   end
 
   def safely_acquire_lock_for_table(table, &block)
+    nested_lock = Thread.current[__method__].present?
```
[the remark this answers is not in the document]
rkrage: Addressed in https://github.com/braintree/pg_ha_migrations/pull/87/commits/bd2c1e92e2cfd7765f579f6baec38bf66f4a1a22

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -553,6 +542,31 @@ def _partitioned_table?(schema, table)
     SQL
   end
 
+  def _partitions_for_table(schema, table, include_sub_partitions: false)
```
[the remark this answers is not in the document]
rkrage: So I'm going back and forth on this one. With the most recent changes, this method accepts a table struct and returns an array of table structs. I guess we could have two methods `_partition_structs_for_table` (or something similar) for use internally, and `partitions_for_table` that would accept a string and return a list of fully qualified child tables.

lib/pg_ha_migrations/struct.rb
```diff
@@ -0,0 +1,59 @@
+module PgHaMigrations::Struct
+  Table = Struct.new(:name, :schema) do
+    delegate :inspect, to: :name
+
+    def fully_qualified_name
+      @fully_qualified_name ||= [
+        ActiveRecord::Base.connection.quote_schema_name(schema),
+        ActiveRecord::Base.connection.quote_table_name(name),
+      ].join(".")
+    end
```
rkrage, line 10: We _could_ add even more methods here, like `partitioned?` and `valid?` (for the index struct below), but I like that these are data objects that only contain light validation and formatting

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -490,14 +482,16 @@ def _quoted_partman_schema
...
+
+    [table.schema, table.name].each do |identifier|
       if identifier.to_s !~ /^[a-z_][a-z_\d]*$/
         raise PgHaMigrations::InvalidMigrationError, "Partman requires schema / table names to be lowercase with underscores"
       end
     end.join(".")
   end
 
-  def _schema_and_table_for(table)
+  def _struct_from_table_name(table)
```
jcoleman, line 494: I think we should move this code directly into `Struct::Table.from_table_name`.

lib/pg_ha_migrations/struct.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    def present?
+      name.present? && schema.present?
+    end
+  end
+
+  Index = Struct.new(:name, :table) do
```
jcoleman, line 17: I think Struct::Table and Struct::Index should maybe be the same thing. Or, at the very least, table should just be Relation, and Index should be a subclass of that.

lib/pg_ha_migrations/safe_statements.rb
rkrage: Related to my comment above (https://github.com/braintree/pg_ha_migrations/pull/87#discussion_r1324759337), I was trying to draw a line with the struct objects such that they only do light validation / formatting and don't actually reach out to the database. If we want to cross that line, I think maybe we'd want to move some of the other helper methods as well like `partitioned?`, `child_partitions`, and `valid?`

lib/pg_ha_migrations/struct.rb
rkrage: Addressed in https://github.com/braintree/pg_ha_migrations/pull/87/commits/bb4b0ce47c6e88de1a0f96381583efa72a8b4495

lib/pg_ha_migrations/safe_statements.rb
rkrage: Update: to hell with the line. This feels pretty clean: https://github.com/braintree/pg_ha_migrations/pull/87/commits/8c05071a751fc560ed02b730f011c22f155527f3

lib/pg_ha_migrations/struct.rb
```diff
@@ -0,0 +1,134 @@
...
+
+      schema = connection.select_value(<<~SQL)
+        SELECT schemaname
+        FROM pg_tables
+        WHERE tablename = #{connection.quote(pg_name.identifier)} AND schemaname = #{schema_conditional}
+        ORDER BY array_position(current_schemas(false), schemaname)
+        LIMIT 1
+      SQL
+
+      raise ArgumentError, "Could not find table #{table}" unless schema.present?
```
rkrage, line 40: This was previously a migration error, but it seems weird to raise that since this class can be used outside of migrations. Maybe we need to add a new error class?

lib/pg_ha_migrations/safe_statements.rb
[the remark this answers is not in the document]
jcoleman: Ah, that's a good point. Take a string _or_ a table struct in would probably be good for a public function, but for the return value I think the struct instances would be good.

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -582,16 +520,35 @@ def exec_migration(conn, direction)
...
+    nested_lock = Thread.current[__method__]
+
     _check_postgres_adapter!
-    table = table.to_s
-    quoted_table_name = connection.quote_table_name(table)
+
+    table = PgHaMigrations::Struct::Table.from_table_name(table)
+
+    # Disallow nested locks unless targeting the same table
+    if nested_lock && nested_lock != table
```
jcoleman, line 530: Now that this is the actual value we can simply:
```suggestion
    if nested_lock != table
```

spec/safe_statements_spec.rb
```diff
@@ -3577,6 +3762,32 @@ def up
...
             end
 
+            it "raises error when attempting nested lock on different table" do
+              ActiveRecord::Base.connection.execute("CREATE TABLE foo(pk SERIAL, i INTEGER)")
+
+              expect do
+                migration.safely_acquire_lock_for_table(table_name) do
+                  expect(locks_for_table(table_name, connection: alternate_connection)).not_to be_empty
+
+                  # attempting a nested lock twice to ensure the
```
jcoleman, line 3772: I like that this is tested.

lib/pg_ha_migrations/struct.rb
jcoleman: If we don't have a need for them now, that's fine, but I wouldn't shy from adding them if it made the code cleaner.

lib/pg_ha_migrations/safe_statements.rb
jcoleman: Yeah, that seems a lot cleaner.

I wasn't going to mention it, but since you brought it up, I don't think the `Struct` nested module adds anything.

lib/pg_ha_migrations/safe_statements.rb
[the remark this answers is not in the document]
rkrage: We could certainly have a public method that takes in a string and delegates to the struct:

```ruby
def partitions_for_table(table)
  PgHaMigrations::Struct::Table.from_table_name(table).partitions.map(&:fully_qualified_name)
end
```

lib/pg_ha_migrations/struct.rb
jcoleman: That's a good point. Probably still worth a PgHaMigrations scoped error class IMO.

lib/pg_ha_migrations/safe_statements.rb
rkrage: I don't think so. In the case where there isn't a nested lock that would turn into `nil != table` which would evaluate to true and we'd incorrectly raise an error

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -501,58 +482,15 @@ def _quoted_partman_schema
   end
 
   def _fully_qualified_table_name_for_partman(table)
```
jcoleman, line 484: I wouldn't worry about this now, but just to put this in the back of your head: this method is probably a candidate for moving to the Relation struct.

lib/pg_ha_migrations/struct.rb
```diff
@@ -0,0 +1,134 @@
...
+
+        tables.concat(sub_partitions)
+      end
+
+      tables
+    end
+  end
+
+  class Index < Relation
+    # TODO: implement shortening to ensure < 63 bytes
```
jcoleman, line 83: Can we add an Issue for this?

spec/safe_statements_spec.rb
```diff
@@ -1,18 +1,26 @@
 require "spec_helper"
 
 RSpec.describe PgHaMigrations::SafeStatements do
-  TableLock = Struct.new(:table, :lock_type, :granted)
+  TableLock = Struct.new(:table, :lock_type, :granted, :pid)
   def locks_for_table(table, connection:)
+    identifiers = table.to_s.split(".")
```
jcoleman, line 6: Will this ever come in with quoting?

spec/safe_statements_spec.rb
```diff
@@ -1,18 +1,26 @@
...
+
     values = connection.execute(<<-SQL)
-      SELECT pg_class.relname AS table, pg_locks.mode AS lock_type, granted
+      SELECT pg_class.relname AS table, pg_locks.mode AS lock_type, granted, pid
       FROM pg_locks
       JOIN pg_class ON pg_locks.relation = pg_class.oid
+      JOIN pg_namespace ON pg_class.relnamespace = pg_namespace.oid
       WHERE pid IS DISTINCT FROM pg_backend_pid()
-        AND pg_class.relkind = 'r'
+        AND pg_class.relkind IN ('r', 'p')
```
jcoleman, line 18:
```suggestion
        AND pg_class.relkind IN ('r', 'p') -- 'r' is a standard table; 'p' is a partition parent
```

spec/safe_statements_spec.rb
```diff
@@ -1,18 +1,26 @@
...
-  TableLock = Struct.new(:table, :lock_type, :granted)
+  TableLock = Struct.new(:table, :lock_type, :granted, :pid)
   def locks_for_table(table, connection:)
+    identifiers = table.to_s.split(".")
+
+    identifiers.prepend("public") if identifiers.size == 1
+
+    schema, table = identifiers
+
     values = connection.execute(<<-SQL)
```
jcoleman, line 12: I think we can now use this?
```suggestion
    values = connection.structs_from_sql(<<-SQL)
```

Note: this could totally be a separate PR.

spec/safe_statements_spec.rb
```diff
@@ -3479,6 +3498,168 @@ def up
               end
             end
 
+            it "does not wait to acquire a lock if same table in different schema is blocked" do
```
jcoleman, line 3501:
```suggestion
            it "does not wait to acquire a lock if a table with the same name but in different schema is blocked" do
```

spec/safe_statements_spec.rb
rkrage: I don't think so, but we could use `PgHaMigrations::Struct::Table.from_table_name(table)` if we wanted. This is just a test helper though so I don't feel too strongly

lib/pg_ha_migrations/safe_statements.rb
rkrage: Addressed in https://github.com/braintree/pg_ha_migrations/pull/87/commits/bf4407d629c2b4eb8e5f1d23f85f9a8087e8fd06

lib/pg_ha_migrations/struct.rb
rkrage: Addressed in https://github.com/braintree/pg_ha_migrations/pull/87/commits/bf4407d629c2b4eb8e5f1d23f85f9a8087e8fd06

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -582,16 +520,35 @@ def exec_migration(conn, direction)
...
+
     _check_postgres_adapter!
-    table = table.to_s
-    quoted_table_name = connection.quote_table_name(table)
+
+    table = PgHaMigrations::Struct::Table.from_table_name(table)
+
+    # Disallow nested locks unless targeting the same table
+    if nested_lock && nested_lock != table
+      raise PgHaMigrations::InvalidMigrationError, "Nested lock detected! Cannot acquire lock on #{table.fully_qualified_name} while #{nested_lock.fully_qualified_name} is locked."
```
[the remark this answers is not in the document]
rkrage: Addressed in https://github.com/braintree/pg_ha_migrations/pull/87/commits/bf4407d629c2b4eb8e5f1d23f85f9a8087e8fd06

spec/safe_statements_spec.rb
rkrage: Addressed in https://github.com/braintree/pg_ha_migrations/pull/87/commits/3d164e2cf3bced19fd3aca90fa360ede80d7c9db

spec/safe_statements_spec.rb
rkrage: Addressed in https://github.com/braintree/pg_ha_migrations/pull/87/commits/bf4407d629c2b4eb8e5f1d23f85f9a8087e8fd06

lib/pg_ha_migrations/safe_statements.rb
jcoleman: Ah, I see that now.

jcoleman (approved):
Let's squash into relevant commits before merging.

PR #88, opened 2023-09-08
Support for different lock modes

rkrage:
Addresses #39


lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -600,19 +600,26 @@ def exec_migration(conn, direction)
     super(conn, direction)
   end
 
-  def safely_acquire_lock_for_table(table, &block)
+  def safely_acquire_lock_for_table(table, mode: :access_exclusive, &block)
```
rkrage, line 603: Is it safe to skip the blocking transactions check if the mode is `access_share` or `row_share`?

spec/lock_mode_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+          access_share
+          row_share
+          row_exclusive
+          share_update_exclusive
+          share
+          share_row_exclusive
+          exclusive
+        ],
+      },
+    }.each do |mode, other_modes|
```
rkrage, line 104: Creating this hash feels kind of dumb, but I kinda like how explicit it is

rkrage: I rewrote the tests to be even more explicit in https://github.com/braintree/pg_ha_migrations/pull/88/commits/e639e54fe2311f2eff38273636af96b023d488d1

lib/pg_ha_migrations/safe_statements.rb
rkrage: Should be addressed in https://github.com/braintree/pg_ha_migrations/pull/88/commits/cd6caf9f57e99319a31fccc87d18d4904ca0b618

lib/pg_ha_migrations/relation.rb
```diff
@@ -77,6 +77,42 @@ def partitions(include_sub_partitions: false)
...
+    def mode=(mode)
+      if mode.present?
+        @mode = LockMode.new(mode)
+      else
+        @mode = mode
+      end
+    end
+
+    def conflicts_with?(other)
+      mode.nil? || other.mode.nil? || mode.conflicts_with?(other.mode)
```
jcoleman, line 104: I think this should verify that the table matches also.

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -546,7 +549,7 @@ def safely_acquire_lock_for_table(table, &block)
         blocking_transactions = PgHaMigrations::BlockingDatabaseTransactions.find_blocking_transactions("#{PgHaMigrations::LOCK_TIMEOUT_SECONDS} seconds")
         blocking_transactions.any? do |query|
           query.tables_with_locks.any? do |table|
-            target_tables.include?(table)
+            target_tables.include?(table) && table_with_lock.conflicts_with?(table)
```
jcoleman, line 552: If we make `conflicts_with?` check table equality then we can drop the `include?` here. Right now can't `table_with_lock` be different from `table`?

I think we should rename the arg `table` now to `locked_table`; there are too many "table" variables in scope now, and I had to read this like 3 times.

Does acquiring a lock on a child table mean you necessarily have a (perhaps lower mode) lock on the parent table? I'm trying to figure out here if we need to check against not the requested table but all target tables, and I'm also trying to understand if we might need to check parent tables (I think not, but I'm having a hard time reasoning about it right now).

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -561,13 +564,13 @@ def safely_acquire_lock_for_table(table, &block)
...
           method(adjust_timeout_method).call(PgHaMigrations::LOCK_TIMEOUT_SECONDS) do
-            connection.execute("LOCK #{table.fully_qualified_name};")
+            connection.execute("LOCK #{table_with_lock.fully_qualified_name} IN #{table_with_lock.mode.to_sql} MODE;")
           end
           successfully_acquired_lock = true
         rescue ActiveRecord::StatementInvalid => e
           if e.message =~ /PG::LockNotAvailable.+ lock timeout/ || e.message =~ /PG::QueryCanceled.+ statement timeout/
             sleep_seconds = PgHaMigrations::LOCK_FAILURE_RETRY_DELAY_MULTLIPLIER * PgHaMigrations::LOCK_TIMEOUT_SECONDS
-            say "Timed out trying to acquire an exclusive lock on the #{table.fully_qualified_name} table."
+            say "Timed out trying to acquire #{table_with_lock.mode.to_sql} lock on the #{table_with_lock.fully_qualified_name} table."
```
jcoleman, line 573: We don't have a/an here anymore, but having that would be hard to do correctly (ActiveSupport probably has something to do it for you though?)

lib/pg_ha_migrations/lock_mode.rb
```diff
@@ -0,0 +1,100 @@
...
+      share_update_exclusive
+      share
+      share_row_exclusive
+      exclusive
+      access_exclusive
+    ]
+
+    attr_reader :mode
+
+    delegate :inspect, :to_s, to: :mode
```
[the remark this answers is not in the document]
rkrage: Looking into that

lib/pg_ha_migrations/relation.rb
```diff
@@ -77,6 +77,42 @@ def partitions(include_sub_partitions: false)
     end
   end
 
+  class TableWithLock < Table
```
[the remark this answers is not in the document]
rkrage: I can try to mess around with that

lib/pg_ha_migrations/relation.rb
```diff
@@ -77,6 +77,42 @@ def partitions(include_sub_partitions: false)
...
+        @mode = mode
+      end
+    end
+
+    def conflicts_with?(other)
+      mode.nil? || other.mode.nil? || mode.conflicts_with?(other.mode)
+    end
+
+    def partitions(include_sub_partitions: false)
+      super.each { |table_with_lock| table_with_lock.mode = mode }
```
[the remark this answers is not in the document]
rkrage: Hmmm if we implement your suggested change to `conflicts_with?`, I think we _do_ need to populate the target lock mode on these partitions... Gonna play around with it

lib/pg_ha_migrations/safe_statements.rb
rkrage: Will have to play around with the `conflicts_with?` change, but definitely agree on the naming. I wrote this code and it's confusing to me so that's probably not a good sign.

> Does acquiring a lock on a child table mean you necessarily have a (perhaps lower mode) lock on the parent table?

I don't think so. Or at least I'm not seeing anything in `pg_locks` that would indicate the parent table gets locked when locking a child partition:

```
pg_ha_migrations_test=# select relation::regclass, locktype, mode from pg_locks;
 relation |  locktype  |      mode
----------+------------+-----------------
 pg_locks | relation   | AccessShareLock
          | virtualxid | ExclusiveLock
(2 rows)

pg_ha_migrations_test=# begin;
BEGIN
pg_ha_migrations_test=*# lock foos3_default;
LOCK TABLE
pg_ha_migrations_test=*# select relation::regclass, locktype, mode from pg_locks;
   relation    |   locktype    |        mode
---------------+---------------+---------------------
 pg_locks      | relation      | AccessShareLock
               | virtualxid    | ExclusiveLock
               | transactionid | ExclusiveLock
 foos3_default | relation      | AccessExclusiveLock
(4 rows)
```

lib/pg_ha_migrations/safe_statements.rb
rkrage: I _could_ look into that, but it honestly reads pretty decently to me as:

> Timed out trying to acquire ACCESS EXCLUSIVE lock on the "public"."foos3" table.

lib/pg_ha_migrations/lock_mode.rb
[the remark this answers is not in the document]
rkrage: I don't think we really gain a lot by making this a struct. We would still want `to_s` and `inspect` delegated to `:mode`:

```ruby
Foo = Struct.new(:mode)
mine = Foo.new(:bar)
puts mine.to_s
#<struct Foo mode=:bar>
puts mine.inspect
#<struct Foo mode=:bar>
```

lib/pg_ha_migrations/relation.rb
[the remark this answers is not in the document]
rkrage: Addressed in https://github.com/braintree/pg_ha_migrations/pull/88/commits/b9cd3000fc10b884c60b0520365cca3e87ac1d41

lib/pg_ha_migrations/relation.rb
rkrage: Addressed in https://github.com/braintree/pg_ha_migrations/pull/88/commits/b9cd3000fc10b884c60b0520365cca3e87ac1d41

lib/pg_ha_migrations/lock_mode.rb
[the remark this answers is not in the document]
jcoleman: I don't think we want to delegate `inspect` to `mode` is part of my point: IMO that breaks standard Ruby expectations.

Nonetheless you still get equality constructs by default with Struct.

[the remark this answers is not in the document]
rkrage: I don't have strong opinions on the `inspect` delegation, but I still don't think we gain much by making this a struct. We need more than equality constructs. We need _comparable_ constructs (is lock mode foo less than lock mode bar)

spec/safe_statements_spec.rb
```diff
@@ -3763,6 +3925,30 @@ def up
               expect(locks_for_table(table_name, connection: alternate_connection)).to be_empty
             end
 
+            it "does not allow re-entrancy when lock escalation detected" do
```
jcoleman, line 3928: I know there's nothing in our usage that would change this, but could we add a test that exiting the inner call does not release the lock?

rkrage: We discussed this offline. Pushed up some more tests in fb6284b

lib/pg_ha_migrations/lock_mode.rb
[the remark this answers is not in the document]
rkrage: Removed the delegation in https://github.com/braintree/pg_ha_migrations/pull/88/commits/ccfe1d23a3317bd090d9dafccc72ec572918f80c

jcoleman (approved):
Please squash for merging.

PR #91, opened 2023-10-10
Add support for Rails 7.1

rkrage:
Also adds Postgres 16 to the build matrix


spec/safe_statements_spec.rb
```diff
@@ -1796,6 +1796,10 @@ def up
           end
 
           it "raises error when child index name is too large" do
+            if ActiveRecord::VERSION::MAJOR >= 7 && ActiveRecord::VERSION::MINOR >= 1
+              skip "Rails 7.1+ will automatically generate index names less than 63 bytes"
+            end
```
rkrage, line 1801: This test will actually be removed entirely in this PR: https://github.com/braintree/pg_ha_migrations/pull/86

So I think it's fine to do this in the short term

jcoleman (approved):

PR #93, opened 2023-11-17
Method to safely add index on empty table

rkrage:
This will raise an error if it detects that the table contains data and/or the total size on disk is greater than 10 megabytes. Otherwise, we safely acquire a `SHARE` lock and run standard `CREATE INDEX ...`


lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -154,6 +154,16 @@ def unsafe_make_column_not_nullable(table, column, options={}) # options arg is
     end
   end
 
+  def safe_add_index_on_empty_table(table, columns, options={})
+    table_struct = PgHaMigrations::Table.from_table_name(table)
+
+    if connection.select_value("SELECT EXISTS (SELECT 1 FROM #{table_struct.fully_qualified_name} LIMIT 1)")
+      raise PgHaMigrations::InvalidMigrationError, "Table #{table_struct.inspect} has rows. Please use safe_add_concurrent_index instead."
+    end
```
rkrage, line 162: So I don't think this is 100% guaranteed to be safe, as the application _could_ write something to the table in the split second after this query runs. But in practice, I think that would be pretty rare

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -204,21 +214,18 @@ def safe_add_concurrent_partitioned_index(
...
+      parent_table.fully_qualified_name,
+      columns,
+      name: parent_index.name,
+      if_not_exists: if_not_exists,
+      using: using,
+      unique: unique,
+      where: where,
+      comment: comment,
+      algorithm: :only, # see lib/pg_ha_migrations/hacks/add_index_on_only.rb
+    )
```
rkrage, line 228: This changed because locking concerns have been moved to the `unsafe_add_index` method itself

lib/pg_ha_migrations/safe_statements.rb
jcoleman: Also, it would have to insert a _ton_ of data to matter.

jcoleman: Actually, what if we re-checked this under the lock? Then it would be guaranteed to be safe.

rkrage: Oh yeah that's a good idea

lib/pg_ha_migrations/unsafe_statements.rb
```diff
@@ -80,7 +80,15 @@ def unsafe_add_index(table, column_names, options = {})
       raise PgHaMigrations::InvalidMigrationError, "ActiveRecord drops the :opclass option when supplying a string containing an expression or list of columns; instead either supply an array of columns or include the opclass in the string for each column"
     end
 
-    execute_ancestor_statement(:add_index, table, column_names, **options)
+    if options[:algorithm] == :concurrently
+      # Cannot be run inside a transaction so we are unable to use
+      # safely_acquire_lock_for_table to take out ShareUpdateExclusiveLock
+      execute_ancestor_statement(:add_index, table, column_names, **options)
```
jcoleman, line 86: This is an interesting catch. This is a project philosophy question, but should we insist that the user call `safe_add_index_concurrently`?

spec/safe_statements_spec.rb
```diff
@@ -1256,6 +1256,59 @@ def up
...
+              end
+            end
+
+            expect do
+              test_migration.suppress_messages { test_migration.migrate(:up) }
+            end.to make_database_queries(matching: /CREATE INDEX "index_foos_on_bar"/, count: 1)
+
+            indexes = ActiveRecord::Base.connection.indexes("foos")
+            expect(indexes.size).to eq(1)
+            expect(indexes.first).to have_attributes(:table => "foos", :name => "index_foos_on_bar", :columns => ["bar"])
```
jcoleman, line 1282: Nit:
```suggestion
            expect(indexes.first).to have_attributes(table: "foos", name: "index_foos_on_bar", columns: ["bar"])
```

lib/pg_ha_migrations/safe_statements.rb
rkrage: With the latest commit  I'm doing locking directly in the `safe_add_index_on_empty_table` method, so I backed this change out for now. I think we can more holistically solve explicit locking for various generic methods in a different PR

lib/pg_ha_migrations/unsafe_statements.rb
rkrage: Not sure I'm following this one

spec/safe_statements_spec.rb
rkrage: Addressed in https://github.com/braintree/pg_ha_migrations/pull/93/commits/e668a5c0311bc26dbfcdfca881ed401d469b8f86

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -506,6 +522,14 @@ def _type_is_enum(type)
     ActiveRecord::Base.connection.select_values("SELECT typname FROM pg_type JOIN pg_enum ON pg_type.oid = pg_enum.enumtypid").include?(type.to_s)
   end
 
+  def _ensure_empty_table!(table)
```
rkrage, line 525: Do you think this would be useful as a public method?

rkrage: If so, we'd need to make the error message a bit more generic

rkrage:
I'm also wondering if maybe we want this method to just be called `safe_add_index`? Or do you like that the validation it's doing is in the name?

lib/pg_ha_migrations/unsafe_statements.rb
jcoleman: This seems to be from an earlier version that updated where we did locking. Since that's no longer true in the PR, I think we can wait on this (but I also don't quite remember the context of my question about project policy, and I'm struggling to find the version of the PR at the time I left the comment, so I can't see the full context of this change...)

lib/pg_ha_migrations/safe_statements.rb
rkrage: Addressed in https://github.com/braintree/pg_ha_migrations/pull/93/commits/058c4d94367fe3679deff0fc8ceacaaa0f79fccd

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -506,6 +522,14 @@ def _type_is_enum(type)
...
+    table = PgHaMigrations::Table.from_table_name(table)
+
+    if connection.select_value("SELECT EXISTS (SELECT 1 FROM #{table.fully_qualified_name} LIMIT 1)")
```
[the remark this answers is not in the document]
rkrage: Addressed in https://github.com/braintree/pg_ha_migrations/pull/93/commits/058c4d94367fe3679deff0fc8ceacaaa0f79fccd

Issue #89, opened 2023-09-17
Ensure that generated index names are <= 63 bytes


rkrage:
The max length for Postgres identifiers is 63 bytes. Recently, Rails introduced a mechanism for hashing index metadata to ensure generated index names are under this limit: https://github.com/rails/rails/pull/47753

It would be really helpful to backport this functionality, particularly to avoid extra validation when creating an index on a sub-partitioned table.

PR #94, opened 2023-11-20
Generate index names less than 63 bytes

rkrage:
This was previously in https://github.com/braintree/pg_ha_migrations/pull/86, but that only solved the problem for creating partitioned indexes. This more holistically addresses #89

Essentially backports [this feature](https://github.com/rails/rails/pull/47753) into older versions of Rails

### DIFF
--- a/lib/pg_ha_migrations/relation.rb
+++ b/lib/pg_ha_migrations/relation.rb
@@ -96,9 +96,23 @@ module PgHaMigrations
   end
 
   class Index < Relation
-    # TODO: implement shortening to ensure < 63 bytes
+    MAX_NAME_SIZE = 63 # bytes
+
     def self.from_table_and_columns(table, columns)
-      new(connection.index_name(table.name, columns), table)
+      name = connection.index_name(table.name, columns)
+
+      # modified from https://github.com/rails/rails/pull/47753
+      if name.bytesize > MAX_NAME_SIZE
+        hashed_identifier = "_#{OpenSSL::Digest::SHA256.hexdigest(name).first(10)}"
+        description = name.sub("index_#{table.name}_on", "idx_on")
+
+        short_limit = MAX_NAME_SIZE - hashed_identifier.bytesize
+        short_description = description.mb_chars.limit(short_limit).to_s
+
+        name = "#{short_description}#{hashed_identifier}"
+      end
+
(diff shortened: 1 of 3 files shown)


lib/pg_ha_migrations/unsafe_statements.rb
```diff
@@ -80,6 +80,16 @@ def unsafe_add_index(table, column_names, options = {})
       raise PgHaMigrations::InvalidMigrationError, "ActiveRecord drops the :opclass option when supplying a string containing an expression or list of columns; instead either supply an array of columns or include the opclass in the string for each column"
     end
 
+    validated_table = PgHaMigrations::Table.from_table_name(table)
```
rkrage, line 83: Subtle, but maybe relevant behavior change here. This will raise a `PgHaMigrations::UndefinedTableError` instead of bubbling up an error from postgres when the table doesn't exist. I _think_ that's okay...

Oh and I think I missed a test for this. Doing that now

lib/pg_ha_migrations/relation.rb
```diff
@@ -96,9 +96,23 @@ def partitions(include_sub_partitions: false, include_self: false)
   end
 
   class Index < Relation
-    # TODO: implement shortening to ensure < 63 bytes
+    MAX_NAME_SIZE = 63 # bytes
+
     def self.from_table_and_columns(table, columns)
```
jcoleman, line 101: This function is always used for constructing a new index name (i.e., as opposed to getting existing indexes)?

rkrage: Currently yes, although the class itself has a `valid?` method which can be used to check for both the existence and validity of an index. This brings up a really good point about the consistency of all of these structs. Like for tables, `from_table_name` expects the table to exist (and will raise an error if it doesn't). And the optional lock mode on a relation can be both an existing lock and a future target lock. It might be worthwhile to refactor all of this to be more consistent, although I don't have good sense of how to do that (and it most definitely should be in a different PR)

lib/pg_ha_migrations/unsafe_statements.rb
jcoleman: That seems fine.

jcoleman (approved):

PR #96, opened 2024-02-28
Add a configuration option for always disabling ddl transactions

ChrisVoxland:
This adds a configuration option: `always_disable_ddl_transactions` to allow folks to easily disable the default behavior of making all migrations run without a DDL transaction.

The reasoning for that as the default makes sense, but I found it violated the principle of least surprise when I was first experimenting with this gem, and it was simple enough to make it configurable. 

The existing warnings for concurrent migrations in a DDL transaction seem safe enough to me, and the tradeoff of potential locking with multiple DDL statements in a single transaction seems worth it for those who want to opt into that behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ When `unsafe_*` migration methods support checks of this type you can bypass the
 
 Similarly we believe the `force: true` option to ActiveRecord's `create_table` method is always unsafe, and therefore we disallow it even when calling `unsafe_create_table`. This option won't be enabled by default until 2.0, but you can opt-in by setting `config.allow_force_create_table = false` [in your configuration initializer](#configuration).
 
-[Running multiple DDL statements inside a transaction acquires exclusive locks on all of the modified objects](https://medium.com/paypal-tech/postgresql-at-scale-database-schema-changes-without-downtime-20d3749ed680#cc22). For that reason, this gem [disables DDL transactions](./lib/pg_ha_migrations/hacks/disable_ddl_transaction.rb) by default. You can change this by setting by setting `config.disable_ddl_transactions_by_default = false` [in your configuration initializer](#configuration).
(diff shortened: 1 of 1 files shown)


rkrage:
@ChrisVoxland can we try to come up with a better name for the config? If I'm understanding this correctly, this setting only affects the behavior of `disable_ddl_transaction` if you don't explicitly define it in your Rails config. So I think I'm mostly getting hung up on the `always` prefix

rkrage:
@ChrisVoxland I think `disable_ddl_transactions_by_default` makes sense

> Also if there is a way to do this without code changes that I'm missing, I'd be happy to put up a PR updating the docs to describe that more specifically instead.

I _thought_ there was already a global config for toggling this behavior in Rails but I don't think that actually exists and it can only be turned off on a per migration basis. So it appears the only way to change the default behavior is by monkey patching the migration class, which is what we're doing here 😄 

So yeah, this PR is definitely a welcome change

jcoleman:
@ChrisVoxland I don't have any issues philosophically with adding this kind of configuration, but I do think we have one thing we need to address to be able to merge: some DDL (e.g., `safe_add_concurrent_index` and `safe_validate_check_constraint` and a few others) can't be run inside a transaction.

I think the simplest solution would be to:
1. Have those methods check if transactions are enabled, and, if so, immediately raise an invalid migration error.
2. Ensure the docs (where you already changed them) note that there are some methods which will require you to turn transactions off if you enable them by default.

rkrage:
I think I agree with @ChrisVoxland on this one.

> Have those methods check if transactions are enabled, and, if so, immediately raise an invalid migration error.

Is it really worth adding extra validation to avoid one extra call to Postgres (that will fail immediately with a pretty descriptive error)?

In both cases, whether we raise the error ourselves or let it bubble up from Postgres, the result will be the same: a rolled back transaction.

@jcoleman what do you think?

rkrage:
> First, it doesn't appear a transaction gets opened at all.

I believe this was just a problem with our test setup. See #101

rkrage:
But back to the original conversation. I'm starting to think that we shouldn't allow this at all...

> If a transaction were to be opened, and lock acquisition didn't succeed on the first attempt, then the whole migration will fail

But if there are blocking transactions, _that_ part of `safely_acquire_lock_for_table` will loop until all long running queries finish. 

So here's a concrete example of one of the risks @jcoleman called out in the docs:

> Acquired locks are held until the end of the transaction.

```ruby
# with disable_ddl_transaction set to false

def up
  safe_make_column_nullable :foo, :column
  safe_make_column_nullable :bar, :column
end
```

We might successfully take out the lock on `foo` and quickly return from the first `safe_make_column_nullable` call, _but_ if there is a long running query against `bar` we could be blocking for a very long time... and therefore `foo` would be locked for a very long time. That seems incredibly dangerous.

Issue #63, opened 2021-09-24
Enforce certain safety features even in unsafe_* by default


jcoleman:
From discussion in #22:
> ...it might well be the case that all of the unsafe variants (where it's relevant) should use the safe lock acquisition. I think that's mostly a project policy discussion, since it means the unsafe variants are increasingly not just proxies for the original Rails method (though we've already gone down that path somewhat with the dependent object checks).

ISTM that what we might want here is to move from 2 layers (`safe_*` and `unsafe_*`) to 3 layers: `safe_*`, `unsafe_*` to prevent clear database level issues (e.g., provide safe locking), and insert a new layer (maybe `raw_*`?) to be a pure dispatch into the original ActiveRecord implementation.

PR #98, opened 2024-05-31
Introduce raw methods for direct dispatch

rkrage:
I view this as the first step to addressing #63 

- raw_* methods added, and are direct dispatch to Rails
- standard Rails methods changed (when `disable_default_migration_methods` is `false`), and are now direct dispatch to Rails
- default config values changed in preparation for 2.0 release

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,19 +42,20 @@ There are two major classes of concerns we try to handle in the API:
 We rename migration methods with prefixes to explicitly denote their safety level:
 
 - `safe_*`: These methods check for both application and database safety concerns, prefer concurrent operations where available, set low lock timeouts where appropriate, and decompose operations into multiple safe steps.
-- `unsafe_*`: These methods are generally a direct dispatch to the native ActiveRecord migration method.
+- `unsafe_*`: Using these methods is a signal that the DDL operation is not necessarily safe for a running application. They include basic safety features like safe lock acquisition and dependent object checking, but otherwise dispatch directly to the native ActiveRecord migration method.
+- `raw_*`: These methods are a direct dispatch to the native ActiveRecord migration method.
 
 Calling the original migration methods without a prefix will raise an error.
 
 The API is designed to be explicit yet remain flexible. There may be situations where invoking the `unsafe_*` method is preferred (or the only option available for definitionally unsafe operations).
 
-While `unsafe_*` methods were historically (through 1.0) pure wrappers for invoking the native ActiveRecord migration method, there is a class of problems that we can't handle easily without breaking that design rule a bit. For example, dropping a column is unsafe from an application perspective, so we make the application safety concerns explicit by using an `unsafe_` prefix. Using `unsafe_remove_column` calls out the need to audit the application to confirm the migration won't break the application. Because there are no safe alternatives we don't define a `safe_remove_column` analogue. However there are still conditions we'd like to assert before dropping a column. For example, dropping an unused column that's used in one or more indexes may be safe from an application perspective, but the cascading drop of the index won't use a `CONCURRENT` operation to drop the dependent indexes and is therefore unsafe from a database perspective.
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -56,17 +56,18 @@ There are two major classes of concerns we try to handle in the API:
...
 
 The API is designed to be explicit yet remain flexible. There may be situations where invoking the `unsafe_*` method is preferred (or the only option available for definitionally unsafe operations).
 
 While `unsafe_*` methods were historically (through 1.0) pure wrappers for invoking the native ActiveRecord migration method, there is a class of problems that we can't handle easily without breaking that design rule a bit. For example, dropping a column is unsafe from an application perspective, so we make the application safety concerns explicit by using an `unsafe_` prefix. Using `unsafe_remove_column` calls out the need to audit the application to confirm the migration won't break the application. Because there are no safe alternatives we don't define a `safe_remove_column` analogue. However there are still conditions we'd like to assert before dropping a column. For example, dropping an unused column that's used in one or more indexes may be safe from an application perspective, but the cascading drop of the index won't use a `CONCURRENT` operation to drop the dependent indexes and is therefore unsafe from a database perspective.
```
jcoleman, line 66:
```suggestion
While `unsafe_*` methods were historically (before 2.0) pure wrappers for invoking the native ActiveRecord migration method, there is a class of problems that we can't handle easily without breaking that design rule a bit. For example, dropping a column is unsafe from an application perspective, so we make the application safety concerns explicit by using an `unsafe_` prefix. Using `unsafe_remove_column` calls out the need to audit the application to confirm the migration won't break the application. Because there are no safe alternatives we don't define a `safe_remove_column` analogue. However there are still conditions we'd like to assert before dropping a column. For example, dropping an unused column that's used in one or more indexes may be safe from an application perspective, but the cascading drop of the index won't use a `CONCURRENT` operation to drop the dependent indexes and is therefore unsafe from a database perspective.
```

README.md
```diff
@@ -56,17 +56,18 @@ There are two major classes of concerns we try to handle in the API:
 We rename migration methods with prefixes denoting their safety level:
 
 - `safe_*`: These methods check for both application and database safety concerns, prefer concurrent operations where available, set low lock timeouts where appropriate, and decompose operations into multiple safe steps.
-- `unsafe_*`: These methods are generally a direct dispatch to the native ActiveRecord migration method.
+- `unsafe_*`: These methods are generally a direct dispatch to the native ActiveRecord migration method, but will sometimes include basic safety features.
```
jcoleman, line 59:
```suggestion
- `unsafe_*`: Using these methods is a signal that the DDL operation is not necessarily safe for a running application. They include basic safety features like safe lock acquisition and dependent object checking, but otherwise dispatch directly to the native ActiveRecord migration method.
```

README.md
```diff
@@ -56,17 +56,18 @@ There are two major classes of concerns we try to handle in the API:
...
 
-When `unsafe_*` migration methods support checks of this type you can bypass the checks by passing an `:allow_dependent_objects` key in the method's `options` hash containing an array of dependent object types you'd like to allow. Until 2.0 none of these checks will run by default, but you can opt-in by setting `config.check_for_dependent_objects = true` [in your configuration initializer](#configuration).
+When `unsafe_*` migration methods support checks of this type you can bypass the checks by passing an `:allow_dependent_objects` key in the method's `options` hash containing an array of dependent object types you'd like to allow. These checks will run by default, but you can opt-out by setting `config.check_for_dependent_objects = false` [in your configuration initializer](#configuration).
```
jcoleman, line 68:
```suggestion
For `unsafe_*` migration methods which support checks of this type you can bypass the checks by passing an `:allow_dependent_objects` key in the method's `options` hash containing an array of dependent object types you'd like to allow. These checks will run by default, but you can opt-out by setting `config.check_for_dependent_objects = false` [in your configuration initializer](#configuration).
```

jcoleman (approved):

PR #99, opened 2024-06-05
Refactor README

jcoleman:
It was getting hard to know where to slot in additional information about our safety features, and that was revealed even more by the upcoming `config.disable_ddl_transactions_by_default` docs, so move things around into clearer headings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ We [disable ActiveRecord's DDL transactions](./lib/pg_ha_migrations/hacks/disabl
   * Each newly acquired lock has its own timeout applied (so total lock time is additive).
   * [Safe lock acquisition](#safely_acquire_lock_for_table) (which is used in each migration method where locks will be acquired) can issue multiple lock attempts on lock timeouts (with sleep delays between attempts).
 
-You can change this by resetting `ActiveRecord::Migration.disable_ddl_transaction` in your application.
+Because of the above issues attempting to re-enable transaction migrations forfeits many of the safety guarantees this library provides and may even break certain functionally. If you'd like to experiment with it anyway you can re-enable transactional migrations by adding `self.disable_ddl_transaction = false` to your migration class definition.
 
 ## Usage
 
--- a/README.md
+++ b/README.md
@@ -30,7 +30,31 @@ Or install it yourself as:
 
     $ gem install pg_ha_migrations
 
-## Usage
+## Migration Safety
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -46,41 +70,48 @@ end
...
-Calling the original migration methods without a prefix will raise an error.
+* [Running multiple DDL statements inside a transaction acquires exclusive locks on all of the modified objects](https://medium.com/paypal-tech/postgresql-at-scale-database-schema-changes-without-downtime-20d3749ed680#cc22).
+* Acquired locks are held until the end of the transaction.
+* Multiple locks creates the possibility of deadlocks.
+* Increased exposure to long waits:
+  * Each newly acquired lock has its own timeout applied (so total lock time is additive).
+  * [Safe lock acquisition](#safely_acquire_lock_for_table) (which is used in each migration method where locks will be acquired) can issue multiple lock attempts on lock timeouts (with sleep delays between attempts).
 
-The API is designed to be explicit yet remain flexible. There may be situations where invoking the `unsafe_*` method is preferred (or the only option available for definitionally unsafe operations).
+You can change this by resetting `ActiveRecord::Migration.disable_ddl_transaction` in your application.
```
jcoleman, line 86: I don't think we have any tests for this, so I'm not sure about leaving it in here, but also it was already present, so I wasn't sure if we should yank it in this change.

rkrage: I did some testing with this and I don't think it even works. `disable_ddl_transaction` is a class instance variable and with the way inheritance works in Ruby, it isn't propagated to individual migration classes that inherit from `ActiveRecord::Migration`.

```ruby
# ruby 3.3.0 (2023-12-25 revision 5124f9ac75) [x86_64-linux]
?> class Parent
?>   class << self
?>     attr_accessor :foo
?>   end
>> end
=> [:foo, :foo=]
>> Parent.foo = "bar"
=> "bar"
>> Parent.foo
=> "bar"
?> class Child < Parent
>> end
=> nil
>> Child.foo
=> nil
```

So until https://github.com/braintree/pg_ha_migrations/pull/96 is merged, I don't think we have a way to globally change the DDL transaction behavior. Of course, you should still be able reset this in each individual migration class with:

```ruby
class MyTransactionalMigration < ActiveRecord::Migration[7.1]
  self.disable_ddl_transaction = false

  def up
    # do some stuff
  end
end
```

rkrage (approved):

PR #100, opened 2024-06-05
Safely acquire lock on multiple tables

rkrage:
This functionality will allow us to improve the safety of operations that take out locks on multiple tables (e.g. adding / removing foreign keys).

I also discovered a bug in the nested lock acquisition logic, so this PR fixes that and adds more tests.

I feel like it's tradition at this point for me to include too many changes in a PR and for @jcoleman to ask me to break it up into smaller commits 😆 (~~I will try to do that tomorrow~~ done).

### DIFF
--- a/lib/pg_ha_migrations/relation.rb
+++ b/lib/pg_ha_migrations/relation.rb
@@ -1,4 +1,8 @@
 module PgHaMigrations
+  # This object represents a pointer to an actual relation in Postgres.
+  # The mode attribute is optional metadata which can represent a lock
+  # that has already been acquired or a potential lock that we are
+  # looking to acquire.
   Relation = Struct.new(:name, :schema, :mode) do
     def self.connection
       ActiveRecord::Base.connection
@@ -13,12 +17,6 @@ module PgHaMigrations
       self.mode = LockMode.new(mode) if mode.present?
     end
 
-    def conflicts_with?(other)
-      self.eql?(other) && (
-        mode.nil? || other.mode.nil? || mode.conflicts_with?(other.mode)
-      )
-    end
-
     def fully_qualified_name
       @fully_qualified_name ||= [
         PG::Connection.quote_ident(schema),
@@ -30,6 +28,16 @@ module PgHaMigrations
       name.present? && schema.present?
     end
 
+    def conflicts_with?(other)
+      self.eql?(other) && (
+        mode.nil? || other.mode.nil? || mode.conflicts_with?(other.mode)
+      )
+    end
+
+    # Some code paths need to compare lock modes, while others just need
+    # to determine if the relation is the same object in Postgres, so
+    # equality here is simply looking at the relation name / schema.
+    # To also compare lock modes, #conflicts_with? is used.
     def eql?(other)
       other.is_a?(Relation) && hash == other.hash
     end
--- a/lib/pg_ha_migrations/relation.rb
+++ b/lib/pg_ha_migrations/relation.rb
@@ -14,7 +14,7 @@ module PgHaMigrations
     end
 
     def conflicts_with?(other)
-      self == other && (
+      self.eql?(other) && (
         mode.nil? || other.mode.nil? || mode.conflicts_with?(other.mode)
       )
     end
@@ -30,8 +30,12 @@ module PgHaMigrations
       name.present? && schema.present?
     end
 
-    def ==(other)
-      other.is_a?(Relation) && name == other.name && schema == other.schema
+    def eql?(other)
+      other.is_a?(Relation) && hash == other.hash
+    end
+
+    def hash
+      [name, schema].hash
     end
   end
 
@@ -164,6 +168,28 @@ module PgHaMigrations
           AND pg_namespace.nspname = #{connection.quote(schema)}
           AND pg_class.relname = #{connection.quote(name)}
       SQL
+    end
+  end
+
+  class TableCollection < Set
+    def self.from_table_names(tables, mode=nil)
+      new(tables) { |table| Table.from_table_name(table, mode) }
+    end
+
+    def mode
+      first&.mode
+    end
+
+    def to_sql
+      map(&:fully_qualified_name).join(", ")
+    end
+
+    def with_partitions
+      tables = flat_map do |table|
+        table.partitions(include_sub_partitions: true, include_self: true)
+      end
+
+      self.class.new(tables)
     end
   end
 end
--- a/lib/pg_ha_migrations/relation.rb
+++ b/lib/pg_ha_migrations/relation.rb
@@ -185,7 +185,7 @@ module PgHaMigrations
     attr_reader :raw_set
 
     delegate :each, to: :raw_set
-    delegate :mode, to: :first, allow_nil: true
+    delegate :mode, to: :first
 
     def self.from_table_names(tables, mode=nil)
       new(tables) { |table| Table.from_table_name(table, mode) }
@@ -193,6 +193,14 @@ module PgHaMigrations
 
     def initialize(tables, &blk)
       @raw_set = tables.map(&blk).to_set
+
+      if raw_set.empty?
+        raise ArgumentError, "Expected a non-empty list of tables"
+      end
+
+      if raw_set.uniq(&:mode).size > 1
+        raise ArgumentError, "Expected all tables in collection to have the same lock mode"
+      end
     end
 
     def subset?(other)
--- a/lib/pg_ha_migrations/relation.rb
+++ b/lib/pg_ha_migrations/relation.rb
@@ -29,7 +29,7 @@ module PgHaMigrations
     end
 
     def conflicts_with?(other)
-      self.eql?(other) && (
+      eql?(other) && (
         mode.nil? || other.mode.nil? || mode.conflicts_with?(other.mode)
       )
     end
@@ -40,6 +40,10 @@ module PgHaMigrations
     # To also compare lock modes, #conflicts_with? is used.
     def eql?(other)
       other.is_a?(Relation) && hash == other.hash
+    end
+
+    def ==(other)
+      eql?(other)
     end
 
     def hash
--- a/lib/pg_ha_migrations/relation.rb
+++ b/lib/pg_ha_migrations/relation.rb
@@ -171,13 +171,24 @@ module PgHaMigrations
     end
   end
 
-  class TableCollection < Set
+  class TableCollection
+    include Enumerable
+
+    attr_reader :raw_set
+
+    delegate :each, to: :raw_set
+    delegate :mode, to: :first, allow_nil: true
+
     def self.from_table_names(tables, mode=nil)
       new(tables) { |table| Table.from_table_name(table, mode) }
     end
 
-    def mode
-      first&.mode
+    def initialize(tables, &blk)
+      @raw_set = tables.map(&blk).to_set
+    end
+
+    def subset?(other)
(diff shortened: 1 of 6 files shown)


lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -528,40 +536,64 @@ def exec_migration(conn, direction)
...
+    end
 
     until successfully_acquired_lock
-      while (
+      loop do
+        # If in a nested context and all of the above checks have passed,
+        # we have already acquired the lock so this check is unnecessary.
+        # In fact, it could actually cause a deadlock if a blocking query
+        # was executed shortly after the initial lock acquisition.
+        break if nested_target_tables
```
rkrage, line 579: This is the bug I was referring to in the PR description

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -528,40 +536,64 @@ def exec_migration(conn, direction)
...
+
         blocking_transactions = PgHaMigrations::BlockingDatabaseTransactions.find_blocking_transactions("#{PgHaMigrations::LOCK_TIMEOUT_SECONDS} seconds")
-        blocking_transactions.any? do |query|
+
+        # Locking a partitioned table will also lock child tables (including sub-partitions),
+        # so we need to check for blocking queries on those tables as well
+        target_tables_for_blocking_transactions = target_tables.flat_map do |target_table|
+          target_table.partitions(include_sub_partitions: true, include_self: true)
+        end
```
rkrage, line 587: I moved this here so that we get a fresh list of tables for every iteration of the loop. I think this gives us extra safety if new partitions are added while this method is executing.

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -570,16 +602,15 @@ def safely_acquire_lock_for_table(table, mode: :access_exclusive, &block)
       end
 
       connection.transaction do
-        adjust_timeout_method = connection.postgresql_version >= 9_03_00 ? :adjust_lock_timeout : :adjust_statement_timeout
         begin
-          method(adjust_timeout_method).call(PgHaMigrations::LOCK_TIMEOUT_SECONDS) do
-            connection.execute("LOCK #{target_table.fully_qualified_name} IN #{target_table.mode.to_sql} MODE;")
+          adjust_statement_timeout(PgHaMigrations::LOCK_TIMEOUT_SECONDS) do
```
rkrage, line 606: Lock timeout applies to each individual relation in the query, while statement timeout applies to the entire query. So, if we were to use lock timeout here, the upper limit for the query timeout would be `LOCK_TIMEOUT_SECONDS * <number of tables being locked>` which seems incorrect

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -3,6 +3,14 @@ def safe_added_columns_without_default_value
     @safe_added_columns_without_default_value ||= []
   end
 
+  # This variable is used to track nested lock acquisition.
+  # Each element is an array of table objects with their lock modes.
+  # The order of the array represents the current call stack,
+  # where the most recent method call is the last element.
+  def safely_acquire_lock_for_table_history
+    @safely_acquire_lock_for_table_history ||= []
+  end
```
rkrage, line 12: This seems way cleaner than using a thread variable. Not sure why I did that in the first place...

lmcicat (approved):

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -3,6 +3,14 @@ def safe_added_columns_without_default_value
     @safe_added_columns_without_default_value ||= []
   end
 
+  # This variable is used to track nested lock acquisition.
+  # Each element is a PgHaMigrations::TableCollection object.
+  # The order of the array represents the current call stack,
+  # where the most recent method call is the last element.
+  def safely_acquire_lock_for_table_history
+    @safely_acquire_lock_for_table_history ||= []
```
jcoleman, line 11: Do we explicitly want to allow nested lock acquisition? If I'm reading the source directly, we didn't explicitly support nested lock acquisition before per se, though we allowed it (as long as it wasn't for the same table).

The nested checks were introduced as a result of #39 to ensure that we didn't upgrade locks, but that was focused on preventing one kind of bug rather than explicitly allowing nested locks. AFAICT it was only incidentally allowed before that.

I can't remember if we need that support for e.g. partitions. But without good justification I'm feeling a bit squeamish about supporting it: it's _very easy_ to get into a bad situation here for multiple reasons:

1. Nested calls have the same problems vis-a-vis lock timeouts as allowing transactional DDL. You end up with nested timeouts, as well, which breaks our guarantees around timing.
2. It's easy to order calls in different ways (even with respect to SQL/DML that's executing externally to the migrations!) that could cause deadlocks.

I'm wondering if we should focus instead on safer APIs (like the multiple table locking one here) that target specific use cases directly.

Note: I'm not sure that the concern about deadlocks (2) is actually prevented by this approach either, because I assume Postgres acquires locks on the relations in order of their presentation here...so maybe that's unavoidable.

lib/pg_ha_migrations/safe_statements.rb
rkrage: While I don't disagree, that will add a bit of complexity, and the extra queries required to track the locks via Postgres will add to the total lock time (maybe not meaningfully) when in a nested lock scenario. I think it's worth exploring, but IMO tracking the locks internally is good enough ™️ since folks aren't generally doing complex multi-threaded operations within a migration. Now I'm wondering if we can somehow prevent multi-threaded access in _all_ of our migration methods... that would make me feel better about tracking this stuff in memory

lib/pg_ha_migrations/safe_statements.rb
rkrage: > Do we explicitly want to allow nested lock acquisition?

I think we do... at least internally. There are a handful of methods (`safe_add_index_on_empty_table` for example) that take out their own lock and then delegate to other methods that _also_ take out that same lock. Of course we can refactor the code to get around scenarios like this, but IMO it makes the code a bit cleaner if we don't have to worry about preventing nested locks

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -528,40 +536,64 @@ def exec_migration(conn, direction)
...
+    end
 
     until successfully_acquired_lock
-      while (
+      loop do
+        # If in a nested context and all of the above checks have passed,
+        # we have already acquired the lock so this check is unnecessary.
+        # In fact, it could actually cause a deadlock if a blocking query
+        # was executed shortly after the initial lock acquisition.
+        break if nested_target_tables
```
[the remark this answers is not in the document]
rkrage: I can play around with moving this outside the loop, but I'm _guessing_ that my intention was to avoid nesting the body of the loop any further by introducing an `if` statement

lib/pg_ha_migrations/relation.rb
```diff
@@ -30,8 +30,12 @@ def present?
       name.present? && schema.present?
     end
 
-    def ==(other)
-      other.is_a?(Relation) && name == other.name && schema == other.schema
+    def eql?(other)
```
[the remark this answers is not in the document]
rkrage: Yeah I mean it's not great, but these table objects are used in different code paths and sometimes represent different things. They can represent a table that is currently locked or a table that we're targeting to be locked and we don't have a way to differentiate the two at the moment. They can also just represent a table without an associated lock (or potential lock). There are different code paths where we just want to know if the tables are the same and don't really care about the lock, and others where we need to check if the locks conflict. I think about these objects as pointers to actual relations in postgres and the lock attribute is just some extra metadata. So it makes sense to me that equivalence in ruby is _just_ comparing the schema / table name in postgres.

lib/pg_ha_migrations/relation.rb
```diff
@@ -152,4 +156,26 @@ def valid?
       SQL
     end
   end
+
+  class TableCollection < Set
```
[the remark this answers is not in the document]
rkrage: Fixed in 9b62643a6fb3bd8a0b027a1d11be551588eaade9

lib/pg_ha_migrations/safe_statements.rb
rkrage: Done

lib/pg_ha_migrations/safe_statements.rb
rkrage: We skip the blocking transaction check but actually do execute the SQL to take out the nested lock, but that's an interesting thought... should we just skip it entirely? 🤔

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -533,40 +541,54 @@ def exec_migration(conn, direction)
...
+        raise PgHaMigrations::InvalidMigrationError,
+          "Lock escalation detected! Cannot change lock level from :#{parent_lock_tables.mode} " \
+          "to :#{target_tables.mode} for #{target_tables.to_sql}."
+      end
+    end
 
     until successfully_acquired_lock
-      while (
+      loop do
+        # If in a nested context and all of the above checks have passed,
```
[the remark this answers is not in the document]
rkrage: This seems related to your previous comment:

> Do we simply skip lock acquisition (as a query, I mean) in that case since we track what we've locked?

Are you suggesting we could short-circuit in a nested lock scenario and just execute the block that was passed in?

lib/pg_ha_migrations/relation.rb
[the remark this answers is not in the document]
rkrage: https://github.com/braintree/pg_ha_migrations/pull/100/commits/c70ed819f5c566213dd1be6fe414e34cdd40e3e5

lib/pg_ha_migrations/safe_statements.rb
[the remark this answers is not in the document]
jcoleman: > This seems related to your previous comment:

I should have clarified what I was continuing the conversation on; it was this one: https://github.com/braintree/pg_ha_migrations/pull/100#discussion_r1640169836

lib/pg_ha_migrations/safe_statements.rb
rkrage: It _could_ be, but I also implemented it here 😄 

https://github.com/braintree/pg_ha_migrations/pull/100/commits/3c4a250f7c79a34b7aa5633c87d2e5036b672ce0

lib/pg_ha_migrations/relation.rb
```diff
@@ -166,4 +178,37 @@ def valid?
...
     end
   end
+
+  class TableCollection
+    include Enumerable
+
+    attr_reader :raw_set
+
+    delegate :each, to: :raw_set
+    delegate :mode, to: :first, allow_nil: true
```
jcoleman, line 188: Can there never be different modes on the different tables?

rkrage: With the ways we are currently interacting with this collection, no. I can probably enforce that in the initializer if you'd like to be a bit more defensive. The only downside is more work for me 😆 . Adding a guard like that would require dedicated unit tests (we exercise these classes currently via functional tests of the methods that use them).

jcoleman: I think it's probably worth being defensive if it's not that difficult.

jcoleman (approved):
There's one outstanding comment thread, but I trust you to finish that, so approving now.

PR #101, opened 2024-06-06
Add tests to verify the behavior of disable_ddl_transaction

rkrage:
With recent PRs / conversations around transactional migrations, I figured it would be a good idea to write some basic tests to verify existing behavior. In other tests, we execute test migrations directly which bypasses the transaction logic in AR.

### DIFF
--- a/spec/pg_ha_migrations_spec.rb
+++ b/spec/pg_ha_migrations_spec.rb
@@ -182,7 +182,10 @@ RSpec.describe PgHaMigrations do
           # Because our method now lives outside of the inheritance
           # hierarchy we can go back to using RSpec's method mocking.
           expect(ActiveRecord::Base.connection).to receive(:pg_ha_migrations_test_method).and_return("sentinel_value")
-          expect(subclass.new.unsafe_pg_ha_migrations_test_method).to eq("sentinel_value")
+
+          subclass.suppress_messages do
+            expect(subclass.new.unsafe_pg_ha_migrations_test_method).to eq("sentinel_value")
+          end
         end
       end
     end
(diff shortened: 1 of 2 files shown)


spec/pg_ha_migrations_spec.rb
```diff
@@ -182,7 +182,10 @@ def pg_ha_migrations_test_method
           # Because our method now lives outside of the inheritance
           # hierarchy we can go back to using RSpec's method mocking.
           expect(ActiveRecord::Base.connection).to receive(:pg_ha_migrations_test_method).and_return("sentinel_value")
-          expect(subclass.new.unsafe_pg_ha_migrations_test_method).to eq("sentinel_value")
+
+          subclass.suppress_messages do
+            expect(subclass.new.unsafe_pg_ha_migrations_test_method).to eq("sentinel_value")
+          end
```
rkrage, line 188: I just noticed a bunch of this in the test output:

```
-- pg_ha_migrations_test_method(nil)
   -> 0.0001s
```

spec/safe_statements_spec.rb
```diff
@@ -100,6 +100,68 @@ def pool_config
...
+
+              raise "boom"
+            end
+          end
+        RUBY
+
+        aggregate_failures do
+          expect do
+            migration_klass.suppress_messages do
+              ActiveRecord::MigrationContext.new(
```
jcoleman, line 119: Can we add comments explaining that our usually way of testing doesn't invoke ActiveRecord's transaction handingling?

jcoleman (commented):
Can you squash into the two meaningful commits after applying the one suggestion?

rkrage:
> Can you squash into the two meaningful commits after applying the one suggestion?

Done and done

jcoleman (approved):

PR #102, opened 2024-06-26
Bug Fix for `unsafe_remove_check_constraint` in certain Rails version 

imsurajmishra:
We identified that for certain rails version ( 6.1+, 7.0 ) , when we use symbol for constraint name, migration fails :
```
> -- remove_check_constraint("test_models", {:name=>:constraint_check_on_test_int_foo})
ArgumentError: Table 'test_models' has no check constraint for {:name=>:constraint_check_on_test_int_foo}
```

the bug is in this code path :
https://github.com/rails/rails/blob/16d8b82d5e2aca4780c5d10b1fe9a90b33a0e84e/activerecord/lib/active_record/connection_adapters/abstract/schema_statements.rb#L1772-L1773

Note: this bug has been fixed in rails 7.1

This PR mainly overriding `unsafe_remove_check_constraint` method and converts `name` arg to `name.to_s`

### DIFF
--- a/lib/pg_ha_migrations/safe_statements.rb
+++ b/lib/pg_ha_migrations/safe_statements.rb
@@ -598,6 +598,12 @@ module PgHaMigrations::SafeStatements
     Thread.current[__method__] = nil unless nested_target_table
   end
 
+  # since rails versions < 7.1 has bug which does not handle symbol for
+  # constraint name, we are converting name to string explicitly to solve that.
+  def raw_remove_check_constraint(table, name:, **options)
+    super(table, name: name.to_s, **options)
+  end
+
   def adjust_lock_timeout(timeout_seconds = PgHaMigrations::LOCK_TIMEOUT_SECONDS, &block)
     _check_postgres_adapter!
     original_timeout = ActiveRecord::Base.value_from_sql("SHOW lock_timeout").sub(/s\Z/, '').to_i * 1000
--- a/spec/safe_statements_spec.rb
+++ b/spec/safe_statements_spec.rb
@@ -1279,6 +1279,42 @@ RSpec.describe PgHaMigrations::SafeStatements do
           end
         end
 
+        describe "raw_remove_check_constraint" do
+          before(:each) do
+            setup_migration = Class.new(migration_klass) do
+              define_method(:up) do
+                unsafe_create_table :foos
+                unsafe_add_column :foos, :bar, :text
+                unsafe_add_check_constraint :foos, "bar IS NOT NULL", :name => :constraint_foo_bar_is_not_null
+              end
+            end
+            setup_migration.suppress_messages { setup_migration.migrate(:up) }
+          end
+
+          ["constraint_foo_bar_is_not_null", :constraint_foo_bar_is_not_null].each do |constraint_name|
+            it "removes the check constraint when passed as a #{constraint_name.is_a?(Symbol) ? "symbol" : "string"}" do
+              migration = Class.new(migration_klass) do
+                # If 'def up' is used instead, we would lose access to the
+                # constraint_name local variable, so we use
+                # 'define_method(:up)'.
+                define_method(:up) do
+                  raw_remove_check_constraint :foos, name: constraint_name
+                end
+              end
+
+              expect do
+                migration.suppress_messages { migration.migrate(:up) }
+              end.to change {
+                ActiveRecord::Base.connection.select_value <<~SQL
+                  SELECT conname
+                  FROM pg_constraint
+                  WHERE conname = 'constraint_foo_bar_is_not_null'
+                SQL
+              }.from("constraint_foo_bar_is_not_null").to(nil)
+            end
+          end
+        end
+
         describe "safe_make_column_nullable" do
           it "removes the not null constraint from the column" do
             migration = Class.new(migration_klass) do


spec/safe_statements_spec.rb
```diff
@@ -1279,6 +1279,24 @@ def up
           end
         end
 
+        describe "unsafe_remove_check_constraint" do
+          ["constraint_foo_bar_is_not_null", :constraint_foo_bar_is_not_null].each do |constraint_name|
+            it "throws error when symbol is used for contraint name" do
```
lmcicat, line 1284:
```suggestion
            it "does not throw an error when symbol is used for constraint name" do
```

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -598,6 +598,12 @@ def safely_acquire_lock_for_table(table, mode: :access_exclusive, &block)
     Thread.current[__method__] = nil unless nested_target_table
   end
 
+  # since rails version 6.0+ & 7.0 has bug which does not handle symbol for
+  # constraint name, we are converting name to string explicitly to solve that.
```
lmcicat, line 602: Instead of specifying exactly 6.0 & 7.0, should it say something like `< 7.1`? I assume this was an issue in all versions prior to the fix in 7.1

spec/safe_statements_spec.rb
imsurajmishra: done https://github.com/braintree/pg_ha_migrations/pull/102/commits/33b712bc8cda3f8f91c4334f72a8b6c5a4849307#diff-c3a8d2986ae1ff24ef5a2cbd56964ac68d7e52190c5a73887fbb5510e45ccb7aR1283

lib/pg_ha_migrations/safe_statements.rb
imsurajmishra: done https://github.com/braintree/pg_ha_migrations/pull/102/commits/33b712bc8cda3f8f91c4334f72a8b6c5a4849307#diff-e246218fba61b707b474c288f842dc0bf0c83436253bf2b49d05216d222f034cR601

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -598,6 +598,12 @@ def safely_acquire_lock_for_table(table, mode: :access_exclusive, &block)
     Thread.current[__method__] = nil unless nested_target_table
   end
 
+  # since rails versions < 7.1 has bug which does not handle symbol for
+  # constraint name, we are converting name to string explicitly to solve that.
+  def unsafe_remove_check_constraint(table, name:, **options)
```
jcoleman, line 603: I think after https://github.com/braintree/pg_ha_migrations/pull/98 we might consider making this an override of `raw_remove_check_constraint` instead of `unsafe_*` since it's effectively back-porting a bug fix as opposed to providing HA safety for DDL.

@rkrage Curious your thoughts.

spec/safe_statements_spec.rb
```diff
@@ -1279,6 +1279,24 @@ def up
...
+                define_method(:up) do
+                  unsafe_create_table :foos
+                  unsafe_add_column :foos, :bar, :text
+                  unsafe_add_check_constraint :foos, "bar IS NOT NULL", :name => :constraint_foo_bar_is_not_null
+                  unsafe_remove_check_constraint :foos, name: constraint_name
+                end
+              end
+              expect do
+                migration.suppress_messages { migration.migrate(:up) }
+              end.not_to raise_error
```
jcoleman, line 1295: Can we add an assertion that the constraint is removed? That might mean making the `unsafe_add_check_constraint` (and prior) DDL part of setup and the assertions about only the remove call.

lmcicat: Updated: https://github.com/braintree/pg_ha_migrations/pull/102/commits/c52eaecdaaf79e21119b3d141f6347dc1b1d67aa

lib/pg_ha_migrations/safe_statements.rb
lmcicat: Definitely agree with overriding `raw_` instead. updated in this commit: https://github.com/braintree/pg_ha_migrations/pull/102/commits/e66dea125bca3668daa81b2991733211766a467f, but if others disagree, we can revert.

Also, not sure if the method & test belong in the `safe_` files they are currently in? The `safe_statements.rb` file has some `unsafe_` method definitions already, so I left it as is for now.

rkrage: Also agree that we should be overriding the raw method, but I think maybe it should go in `unsafe_statements.rb`. @jcoleman is there a reason we sometimes have unsafe methods in the safe statements file? I generally put safe and unsafe variants of _custom_ methods in the safe file, just so the code is close together. But for basic wrappers of existing Rails methods... maybe the unsafe file is more correct.

jcoleman: I think it might have something to do with the module prepending order, but without testing it I don't remember details. Possible there's a commit message explaining it somewhere.

rkrage: Okay, then I think this is fine for now. I might try to follow up with a PR to move all these methods into more "correct" files and we'll see what tests break.

spec/safe_statements_spec.rb
```diff
@@ -1279,6 +1279,41 @@ def up
...
+              migration = Class.new(migration_klass) do
+                define_method(:up) do
+                  raw_remove_check_constraint :foos, name: constraint_name
+                end
+              end
+
+              RSpec::Matchers.define_negated_matcher :not_raise_error, :raise_error
+              expect do
+                migration.suppress_messages { migration.migrate(:up) }
+              end.to not_raise_error
```
rkrage, line 1305: Do we need to explicitly check if it doesn't raise an error? It would be _slightly_ less explicit in the output, but if we remove this expectation and it does raise an error, the test should still fail, right?

lib/pg_ha_migrations/safe_statements.rb
jcoleman: Oh, I think `unsafe_rename_enum_value` is there simply because there isn't an analogue in ActiveRecord.

But the other ones -- like `unsafe_add_column` I think are there because they reference `safely_acquire_lock_for_table` which is in that same file (and it'd be weird for a module higher up there ancestor chain to call a method at a lower level). We could move that function though.

spec/safe_statements_spec.rb
jcoleman, line 1305: We wouldn't want to define this inside a test, but we could also avoid defining it entirely by simply having nested `expect do` blocks (one for `not_to raise_error` and one for `to change`).

lmcicat: Yeah, that would work! @rkrage also suggested not necessarily needing the `not_raise_error` check so we could just update this expect to only checking the constraint is removed

jcoleman: Yep, that seems fine too.

lib/pg_ha_migrations/safe_statements.rb
rkrage: Maybe just move the utility methods to separate module that gets prepended higher up the chain then?

spec/safe_statements_spec.rb
```diff
@@ -1279,6 +1279,39 @@ def up
           end
         end
 
+        describe "raw_remove_check_constraint" do
+          before(:each) do
+            setup_migration = Class.new(migration_klass) do
+              define_method(:up) do
```
rkrage, line 1285: Nitpick: I think this could just be `def up`

rkrage (approved):

spec/safe_statements_spec.rb
lmcicat: yeah, good point! Removed the check here: https://github.com/braintree/pg_ha_migrations/pull/102/commits/bbc9ac5abe41bd0c1c5aafcd90ab1c559fdb22ec

lmcicat: updated to not have the not_raise_error check: https://github.com/braintree/pg_ha_migrations/pull/102/commits/bbc9ac5abe41bd0c1c5aafcd90ab1c559fdb22ec

spec/safe_statements_spec.rb
lmcicat: if we change it to be `def up` we don't have access to the `constraint_name` local variable anymore.

lib/pg_ha_migrations/safe_statements.rb
jcoleman: Yeah, we could probably do that. Separate work/PR though?

spec/safe_statements_spec.rb
jcoleman: Let's add a comment explaining that.

rkrage: Ah that makes sense. And a comment makes sense too.

I've abused class attributes to get around these scoping problems when doing dynamic class stuff: https://github.com/braintree/pg_ha_migrations/blob/77c3addf776173f113a28139df5e3141a49b74db/spec/safe_statements_spec.rb#L44-L78

But what you've done is better

lmcicat: Added a comment!

jcoleman (approved):

Issue #109, opened 2024-11-25
Rails 8 support


rkrage:
I _think_ the only blocker is [relation_to_struct](https://github.com/jcoleman/relation_to_struct) at the moment, but other code changes may be needed.

rkrage:
Required `relation_to_struct` changes here: https://github.com/jcoleman/relation_to_struct/pull/23

PR #111, opened 2024-11-25
Add support for Rails 8.0

rkrage:
Also:

- Remove Ruby 3.1 from build matrix
- Use `~>` syntax in appraisal definitions such that we pull in the latest patch versions of Rails when the tests run

~~We'll want https://github.com/jcoleman/relation_to_struct/pull/23 to be merged and released first~~

Closes https://github.com/braintree/pg_ha_migrations/issues/109

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,13 +10,13 @@ jobs:
           - 15
           - 16
         ruby:
-          - "3.1"
           - "3.2"
           - "3.3"
         gemfile:
           - rails_7.0
           - rails_7.1
           - rails_7.2
+          - rails_8.0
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -10,13 +10,13 @@ jobs:
           - 15
           - 16
         ruby:
-          - "3.1"
```
rkrage, line 13: Same comment from here: https://github.com/jcoleman/relation_to_struct/pull/23#discussion_r1857340361

The other versions of Rails still support 3.1 but it was much easier to just remove it completely instead of special-casing Rails 8. I think this should be fine as Ruby 3.1 is EOL in 4 months?

jcoleman (approved):

PR #112, opened 2024-11-26
Improve build performance

rkrage:
This diff is basically impossible to view, but the TLDR is that I moved the following utility methods out of the `PgHaMigrations::AllowedVersions::ALLOWED_VERSIONS` loop:

- `safely_acquire_lock_for_table`
- `adjust_lock_timeout`
- `adjust_statement_timeout`
- `ensure_small_table`

These methods do not exercise any of the backwards compatibility features of migrations, so we shouldn't need to run the tests for each migration version. The `safely_acquire_lock_for_table` tests are very slow, so this should hopefully speed things up significantly.

### DIFF
--- a/spec/safe_statements_spec.rb
+++ b/spec/safe_statements_spec.rb
@@ -3957,891 +3957,6 @@ RSpec.describe PgHaMigrations::SafeStatements do
           end
         end
 
-        describe "#adjust_lock_timeout" do
-          let(:table_name) { "bogus_table" }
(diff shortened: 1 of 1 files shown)


rkrage:
Before:

Screenshot 2024-11-26 at 10 31 34 AM

After:

Screenshot 2024-11-26 at 10 31 41 AM

jcoleman (approved):

PR #113, opened 2024-11-27
Implement safe locking in unsafe_* methods when appropriate

rkrage:
Closes https://github.com/braintree/pg_ha_migrations/issues/63

Leaving the below context, but just a note that most of that refactoring was actually done in: https://github.com/braintree/pg_ha_migrations/pull/123

Okay, so this is a massive one, but the diff is so large because I used this as an opportunity to reorganize the tests and moved all of the `unsafe_*` methods into `unsafe_statements.rb`. It became incredibly difficult to locate and modify tests to add locking expectations (and surprise some of those methods didn't even have tests), so I attempted to organize the tests into four categories:

- raw methods that are simply delegated to the underlying Rails method via `delegate_raw_method_to_migration_base_class`
- unsafe methods that are simply delegated to underlying Rails method via `delegate_unsafe_method_to_migration_base_class_with_lock` or `delegate_unsafe_method_to_migration_base_class`
- unsafe methods that we define ourselves
- safe methods that we define ourselves

This additionally removes support for Rails 7.0 (in a separate commit) which gets rid of a fifth type of method:
- raw methods that we define ourselves


lib/pg_ha_migrations/unsafe_statements.rb
```diff
@@ -127,10 +210,6 @@ def unsafe_partman_update_config(table, **options)
 
     config = PgHaMigrations::PartmanConfig.find(_fully_qualified_table_name_for_partman(table))
 
-    if !options[:automatic_maintenance].nil?
-      options[:automatic_maintenance] = options[:automatic_maintenance] ? "on" : "off"
-    end
```
rkrage, line 132: I think this snuck in a previous PR: https://github.com/braintree/pg_ha_migrations/pull/117

I started to add this feature and realized a) this isn't something folks would be doing regularly and b) this was outside the scope of the PR. I guess I never fully cleaned up the changes, so doing so now.

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -622,12 +594,6 @@ def safely_acquire_lock_for_table(*tables, mode: :access_exclusive, &block)
     end
   end
 
-  # since rails versions < 7.1 has bug which does not handle symbol for
-  # constraint name, we are converting name to string explicitly to solve that.
-  def raw_remove_check_constraint(table, name:, **options)
-    super(table, name: name.to_s, **options)
-  end
```
rkrage, line 629: As I was looking to organize methods and their tests, I discovered this unicorn. Rails 7.0 is EOL, so my solution was to just remove this method and Rails 7.0 from our gemspec / build matrix

lib/pg_ha_migrations/unsafe_statements.rb
```diff
@@ -91,15 +106,21 @@ def self.delegate_raw_method_to_migration_base_class(method_name)
   disable_or_delegate_default_method :rename_column, ":rename_column is NOT SAFE! Explicitly call :unsafe_rename_column to proceed"
   disable_or_delegate_default_method :rename_table, ":rename_table is NOT SAFE! Explicitly call :unsafe_rename_table to proceed"
 
-  def unsafe_create_table(table, options={}, &block)
+  # Note that unsafe_* methods defined below do not run dependent object checks
+
+  def unsafe_change_table(*args, &block)
+    raise PgHaMigrations::UnsafeMigrationError.new(":change_table is too generic to even allow an unsafe variant. Use a combination of safe and explicit unsafe migration methods instead")
+  end
```
rkrage, line 113: I don't think we've ever done something like this before, so open to opinions

lib/pg_ha_migrations/unsafe_statements.rb
```diff
@@ -115,7 +136,69 @@ def unsafe_add_index(table, column_names, options = {})
...
+    if ActiveRecord::Base.connection.postgresql_version < 10_00_00
+      raise PgHaMigrations::InvalidMigrationError, "Renaming an enum value is not supported on Postgres databases before version 10"
+    end
+
+    raw_execute("ALTER TYPE #{PG::Connection.quote_ident(name.to_s)} RENAME VALUE '#{PG::Connection.escape_string(old_value)}' TO '#{PG::Connection.escape_string(new_value)}'")
+  end
+
+  def unsafe_make_column_not_nullable(table, column, **options) # options arg is only present for backwards compatiblity
+    safely_acquire_lock_for_table(table) do
+      raw_execute("ALTER TABLE #{table} ALTER COLUMN #{column} SET NOT NULL") # TODO: quoting
```
jcoleman: Is the TODO here new?

jcoleman (commented):
Do we want to have unsafe_statements_spec.rb now?

I left a few comments on the code, but the spec file diff is unreadable; I think the issue is that the removal of something deprecated/defaults is getting confused with the move of the unsafe methods testing. But that no tonly means that the things it thinks are changes aren’t necessarily changes, it means all of the moved stuff is just shown as all net new…Maybe the answer is a commit that moves the tests around in the file (like a pure refactor) and then a separate commit with actual changes?

lib/pg_ha_migrations/unsafe_statements.rb
rkrage: It was, but I pushed up this commit to address (that was fixed in a previous PR and I messed up the rebase)

https://github.com/braintree/pg_ha_migrations/pull/113/commits/9adbc2bfda45636ac7aeb2335ed53bff3aac3261

jcoleman:
I think the diff will be actually readable if rebased on https://github.com/braintree/pg_ha_migrations/pull/123

spec/safe_statements_spec.rb
```diff
@@ -4020,7 +4163,13 @@ def up
...
               end
 
-              setup_migration.suppress_messages { setup_migration.migrate(:up) }
+              expect do
+                migration.suppress_messages { migration.migrate(:up) }
+              end.to raise_error(PgHaMigrations::UndefinedTableError, "Table \"foos3_template\" does not exist in search path")
+            end
+
+            it "raises error when template table does not exist and fully qualified name provided" do
+              create_range_partitioned_table(:foos3, migration_klass)
```
rkrage, line 4172: Unsure why this block is showing up in the diff...

lib/pg_ha_migrations/unsafe_statements.rb
```diff
@@ -29,30 +29,45 @@ def self.delegate_unsafe_method_to_migration_base_class(method_name)
...
-  delegate_unsafe_method_to_migration_base_class :add_check_constraint
-  delegate_unsafe_method_to_migration_base_class :add_column
-  delegate_unsafe_method_to_migration_base_class :add_foreign_key
-  delegate_unsafe_method_to_migration_base_class :change_column
-  delegate_unsafe_method_to_migration_base_class :change_column_default
-  delegate_unsafe_method_to_migration_base_class :change_table
+  # Direct dispatch to underlying Rails method as unsafe_<method_name> with dependent object check / safe lock acquisition
+  delegate_unsafe_method_to_migration_base_class_with_lock :add_check_constraint
+  delegate_unsafe_method_to_migration_base_class_with_lock :add_column
+  delegate_unsafe_method_to_migration_base_class_with_lock :change_column
```
jcoleman, line 55: It seems like we should treat this like we do `change_table`?

spec/safe_statements_spec.rb
```diff
@@ -1841,7 +2036,7 @@ def up
             end.not_to raise_error
           end
 
-          describe "when not configured to disallow two-step new column and adding default" do
+          context "when not configured to disallow two-step new column and adding default" do
```
jcoleman, line 2039: Disapproving stare.

lib/pg_ha_migrations/unsafe_statements.rb
```diff
@@ -29,30 +29,45 @@ def self.delegate_unsafe_method_to_migration_base_class(method_name)
...
+  delegate_unsafe_method_to_migration_base_class_with_lock :change_column_default
+  delegate_unsafe_method_to_migration_base_class_with_lock :remove_check_constraint
+  delegate_unsafe_method_to_migration_base_class_with_lock :remove_column
+  delegate_unsafe_method_to_migration_base_class_with_lock :rename_column
+
+  # Direct dispatch to underlying Rails method as unsafe_<method_name> with dependent object check / no locking
+  #
+  # We don't do locking on execute because it's so generic.
+  # For drop_table and rename_table, it's implied that the table is unused.
```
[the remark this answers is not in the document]
jcoleman: > If we added locking to these methods, I think (5) would fail with a "nested lock detected" error.

It'd fail in development though. Since we support locking multiple tables with `safely_acquire_lock_for_table` I think it'd be easy to fix, right?

Basically my thought comes down to this: we want to be safe by default, and safely acquiring a lock on something that's safe to drop is safe, but failing to acquire that lock safely opens us to the possibility that a mistaken table drop _also_ locks the system up.

lib/pg_ha_migrations/unsafe_statements.rb
jcoleman: I think it's fine, but we should be sure it's documented.

lib/pg_ha_migrations/unsafe_statements.rb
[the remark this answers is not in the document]
rkrage: > I think it'd be easy to fix, right?

Sure but then you would have to move the table creation outside the lock. Can't lock a table that doesn't exist!

But yeah, I see your point. I don't feel too strongly about all this so I'll go ahead and make this change.

A somewhat related question... would it add safety to have some kind of timeout / retry limit for lock acquisition? As it stands `safely_acquire_lock_for_table` just keeps trying indefinitely, correct?

spec/safe_statements_spec.rb
rkrage: 😆  Are you in the camp of just using `describe` everywhere? Really don't feel strongly about this either so I can change it back

jcoleman: Yeah, I use `describe` everywhere for simplicity. I find the describe/context thing a significant bike-shed debate with no clear boundaries.

lib/pg_ha_migrations/unsafe_statements.rb
jcoleman: > Like if you wanted to change a column type would you be forced to use `raw_change_column`?

Hmm, that's a good point. We probably should have a way to do those things discretely (and that would allow a safe version for types, also, since some are binary compatible).

But the non-existence of those now suggests we can punt that to an issue to track it.

rkrage: Yeah I was thinking about that a bit too. I can open an issue and we can discuss further, but I'm assuming the list of binary compatible types needs to be whittled down to type changes that Rails can transparently handle. For example, json -> text appears to be binary compatible, but I imagine Rails might not like that

lib/pg_ha_migrations/unsafe_statements.rb
rkrage: Done in https://github.com/braintree/pg_ha_migrations/pull/113/commits/bee6f73904f3fad69572422813c378d51a2d1140

spec/safe_statements_spec.rb
rkrage: Done in https://github.com/braintree/pg_ha_migrations/pull/113/commits/4815e0476f6d3d3815ac5365218db26d4b4aacea

lib/pg_ha_migrations/unsafe_statements.rb
jcoleman: Yeah, sounds good.

lib/pg_ha_migrations/unsafe_statements.rb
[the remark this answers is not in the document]
jcoleman: > A somewhat related question... would it add safety to have some kind of timeout / retry limit for lock acquisition? As it stands safely_acquire_lock_for_table just keeps trying indefinitely, correct?

Yeah, I think so. That sounds like a good candidate for an issue to discuss further.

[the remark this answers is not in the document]
rkrage: https://github.com/braintree/pg_ha_migrations/issues/127

lib/pg_ha_migrations/unsafe_statements.rb
rkrage: https://github.com/braintree/pg_ha_migrations/issues/128

Issue #106, opened 2024-11-06
safe_make_column_not_nullable


alexspeller:
I think it would be possible to add a safe_make_column_not_nullable method, [source](https://dba.stackexchange.com/a/268128)

I think something like:

```rb

  def safe_make_column_not_nullable(table, column)
    unsafe_execute "ALTER TABLE #{table} ADD CONSTRAINT tmp_not_null_constraint CHECK (#{column} IS NOT NULL) NOT VALID"
    unsafe_execute "ALTER TABLE #{table} VALIDATE CONSTRAINT tmp_not_null_constraint"
    unsafe_execute "ALTER TABLE #{table} ALTER COLUMN #{column} SET NOT NULL"
    unsafe_execute "ALTER TABLE #{table} DROP CONSTRAINT tmp_not_null_constraint"
  end
```

rkrage:
Closed in https://github.com/braintree/pg_ha_migrations/pull/116

PR #114, opened 2024-12-26
Add safe_make_column_not_nullable

alexspeller:
Addresses #106

I also added some quoting for some related methods where it seemed to be missing

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,6 +191,9 @@ Safely make the column not nullable - adds a temporary constraint and uses that 
 safe_make_column_not_nullable :table, :column
 ```
 
+Note:
+This method runs multiple DDL statements non-transactionally. Validating the constraint could fail. In such cases an exception will be raised, and an INVALID constraint will be left on the table.
+
 #### unsafe\_make\_column\_not\_nullable
 
 Unsafely make a column not nullable.
--- a/README.md
+++ b/README.md
@@ -185,7 +185,7 @@ safe_make_column_nullable :table, :column
 ```
 #### safe\_make\_column\_not\_nullable
 
-Safely make the column not nullable - adds a temporary constraint and uses that constraint to proove no values are null before altering the column, then removes the temporary constraint
+Safely make the column not nullable - adds a temporary constraint and uses that constraint to validate no values are null before altering the column, then removes the temporary constraint
 
 ```ruby
 safe_make_column_not_nullable :table, :column
--- a/README.md
+++ b/README.md
@@ -183,6 +183,13 @@ Safely make the column nullable.
 ```ruby
 safe_make_column_nullable :table, :column
 ```
+#### safe\_make\_column\_not\_nullable
+
+Safely make the column not nullable - adds a temporary constraint and uses that constraint to proove no values are null before altering the column, then removes the temporary constraint
+
+```ruby
+safe_make_column_not_nullable :table, :column
+```
 
 #### unsafe\_make\_column\_not\_nullable
 
--- a/lib/pg_ha_migrations/safe_statements.rb
+++ b/lib/pg_ha_migrations/safe_statements.rb
@@ -133,14 +133,44 @@ module PgHaMigrations::SafeStatements
   end
 
   def safe_make_column_nullable(table, column)
+    quoted_table_name = connection.quote_table_name(table)
+    quoted_column_name = connection.quote_column_name(column)
+
     safely_acquire_lock_for_table(table) do
-      raw_execute "ALTER TABLE #{table} ALTER COLUMN #{column} DROP NOT NULL"
+      raw_execute "ALTER TABLE #{quoted_table_name} ALTER COLUMN #{quoted_column_name} DROP NOT NULL"
     end
   end
 
+  # Postgres 12+ can use a valid CHECK constraint to proove that no values of a column are null, avoiding
+  # locking the table when altering a column to NOT NULL
+  #
+  # Source:
+  # https://dba.stackexchange.com/questions/267947/how-can-i-set-a-column-to-not-null-without-locking-the-table-during-a-table-scan/268128#268128
+  # (https://archive.is/X55up)
+  def safe_make_column_not_nullable(table, column)
+    if ActiveRecord::Base.connection.postgresql_version < 12_00_00
+      raise PgHaMigrations::InvalidMigrationError, "Cannot safely make a column non-nullable before Postgres 12"
+    end
+
+    quoted_table_name = connection.quote_table_name(table)
+    quoted_column_name = connection.quote_column_name(column)
+
+    tmp_constraint_name = :tmp_not_null_constraint
+
+    safe_add_unvalidated_check_constraint(table, "#{quoted_column_name} IS NOT NULL", name: tmp_constraint_name)
+    safe_validate_check_constraint(table, name: tmp_constraint_name)
+
+    unsafe_make_column_not_nullable(table, column)
+
+    raw_execute "ALTER TABLE #{quoted_table_name} DROP CONSTRAINT #{tmp_constraint_name}"
+  end
+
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -183,6 +183,13 @@ Safely make the column nullable.
 ```ruby
 safe_make_column_nullable :table, :column
 ```
+#### safe\_make\_column\_not\_nullable
+
+Safely make the column not nullable - adds a temporary constraint and uses that constraint to proove no values are null before altering the column, then removes the temporary constraint
```
jcoleman, line 188:
```suggestion
Safely make the column not nullable - adds a temporary constraint and uses that constraint to validate no values are null before altering the column, then removes the temporary constraint
```

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -133,14 +133,44 @@ def safe_change_column_default(table_name, column_name, default_value)
...
+    quoted_table_name = connection.quote_table_name(table)
+    quoted_column_name = connection.quote_column_name(column)
+
     safely_acquire_lock_for_table(table) do
-      raw_execute "ALTER TABLE #{table} ALTER COLUMN #{column} DROP NOT NULL"
+      raw_execute "ALTER TABLE #{quoted_table_name} ALTER COLUMN #{quoted_column_name} DROP NOT NULL"
     end
   end
 
+  # Postgres 12+ can use a valid CHECK constraint to proove that no values of a column are null, avoiding
```
jcoleman, line 144:
```suggestion
  # Postgres 12+ can use a valid CHECK constraint to validate that no values of a column are null, avoiding
```

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -133,14 +133,44 @@ def safe_change_column_default(table_name, column_name, default_value)
...
+    quoted_column_name = connection.quote_column_name(column)
+
+    tmp_constraint_name = :tmp_not_null_constraint
+
+    safe_add_unvalidated_check_constraint(table, "#{quoted_column_name} IS NOT NULL", name: tmp_constraint_name)
+    safe_validate_check_constraint(table, name: tmp_constraint_name)
+
+    unsafe_make_column_not_nullable(table, column)
+
+    raw_execute "ALTER TABLE #{quoted_table_name} DROP CONSTRAINT #{tmp_constraint_name}"
```
jcoleman, line 165: Could this use `unsafe_remove_constraint`?

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -133,14 +133,44 @@ def safe_change_column_default(table_name, column_name, default_value)
...
+  end
+
   def unsafe_make_column_not_nullable(table, column, options={}) # options arg is only present for backwards compatiblity
+    quoted_table_name = connection.quote_table_name(table)
```
jcoleman, line 169: Thanks for this extra improvement!

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -133,14 +133,44 @@ def safe_change_column_default(table_name, column_name, default_value)
...
+  # (https://archive.is/X55up)
+  def safe_make_column_not_nullable(table, column)
+    if ActiveRecord::Base.connection.postgresql_version < 12_00_00
+      raise PgHaMigrations::InvalidMigrationError, "Cannot safely make a column non-nullable before Postgres 12"
+    end
+
+    quoted_table_name = connection.quote_table_name(table)
+    quoted_column_name = connection.quote_column_name(column)
+
+    tmp_constraint_name = :tmp_not_null_constraint
```
jcoleman, line 158: Are constraint names global? If so, do we want to add a random tag on the end of this?

spec/safe_statements_spec.rb
```diff
@@ -1343,6 +1343,23 @@ def up
...
+            migration = Class.new(migration_klass) do
+              def up
+                unsafe_create_table :foos do |t|
+                  t.text :bar, :null => false
+                end
+                safe_make_column_not_nullable :foos, :bar
+              end
+            end
+
+            migration.suppress_messages { migration.migrate(:up) }
```
jcoleman, line 1357: I wonder if we should add an assertion that this doesn't use locks beyond a certain level.

@rkrage Thoughts on this? I don't think we have this, but I'd love to have a test helper something like this:

```
expect do
  ...
end.not_to acquire_locks_on_table(...).higher_than(...)
```

Or similar.

Thinking out loud here:

I don't think there's a way to get Postgres to log all locks acquired, and so we'd probably have to do something like lock the table and then run the second query to verify what type is waiting in `pg_locks`. That doesn't work easily for multiple statements, though, so I think in this case we might lock the table at a lower level and expect this to proceed. But I think the `ALTER TABLE` will still need an AccessExclusiveLock, it will just be very short lived.

jcoleman (commented):
I'm wondering if we want to also (whether it's in this PR or not) add a method that converts a validated check constraint to a regular `NOT NULL` constraint (the benefit basically being that it would assert that the constraint exists and is already validated).

lib/pg_ha_migrations/safe_statements.rb
rkrage: Yeah, I think constraint names need to be unique within a schema. Because this method is doing multiple things non-transactionally, there is a possibility we could error out before the constraint is removed. So a) I agree with @jcoleman that we should add some random characters to this string and b) we should document that this might leave behind a constraint on failure, similar to creating a [partitioned index](https://github.com/braintree/pg_ha_migrations?tab=readme-ov-file#safe_add_concurrent_partitioned_index).

spec/safe_statements_spec.rb
rkrage: I'll have to think about this a bit, but I think it's going to be super difficult to make a generic helper. As you pointed out, we could acquire an `AccessExclusiveLock` and sleep for x seconds in a thread and then concurrently run the code under test. That code would block until the sleep expires and then we could go inspect the logs for the output from `log_lock_waits`. However, that breaks down for methods that execute multiple SQL statements.

I guess we could do something crazy and mock the `connection#execute` method with `and_wrap_original` such that it automatically does the AEL shenanigans described above before _any_ SQL is executed

rkrage: Since we don't currently have a precedent for writing such a test, I'm not sure if it's fair to put this on @alexspeller 😄 

This can maybe be something we do as a follow on to this PR

lib/pg_ha_migrations/safe_statements.rb
alexspeller: https://www.postgresql.org/docs/current/sql-set-constraints.html#:~:text=Because%20PostgreSQL%20does%20not%20require,will%20act%20on%20all%20matches.

spec/safe_statements_spec.rb
alexspeller: Yeah realistically I don't have the bandwidth at the moment to deep dive on that unfortunately - I didn't see any precedent for more in-depth lock testing for similar methods so I kept it quite basic

jcoleman:
@alexspeller Apologies, for some reason I'm not getting notifications on this repo anymore; I'll take a look at this again.

spec/safe_statements_spec.rb
jcoleman: Yeah, I think we're fine for now; I wanted to at least kick off the conversation internally if nothing else.

jcoleman:
@alexspeller I think @rkrage are going to add wrapping this PR up to our internal development list, and after we wrap it up, we'll merge.

rkrage:
Closing in favor of https://github.com/braintree/pg_ha_migrations/pull/116

PR #116, opened 2025-02-28
Add safe make column not nullable

rkrage:
This is just taking over the previous PR (https://github.com/braintree/pg_ha_migrations/pull/114) to make a few final tweaks.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,6 +1,7 @@
 require "bundler/gem_tasks"
 require "rspec/core/rake_task"
 require "appraisal"
+# In Rails 6 this isn't required in the right order and worked by accident; fixed in rails@0f5e7a66143
 require "logger"
 require_relative File.join("lib", "pg_ha_migrations")
 
--- a/Rakefile
+++ b/Rakefile
@@ -1,6 +1,7 @@
 require "bundler/gem_tasks"
 require "rspec/core/rake_task"
 require "appraisal"
+require "logger"
 require_relative File.join("lib", "pg_ha_migrations")
 
 RSpec::Core::RakeTask.new(:spec)
--- a/lib/pg_ha_migrations/safe_statements.rb
+++ b/lib/pg_ha_migrations/safe_statements.rb
@@ -53,7 +53,7 @@ module PgHaMigrations::SafeStatements
         raise PgHaMigrations::UnsafeMigrationError.new(":default is not safe if the default value is volatile. Use safe_change_column_default afterwards then backfill the data to prevent locking the table")
       end
     elsif options[:null] == false
-      raise PgHaMigrations::UnsafeMigrationError.new(":null => false is NOT SAFE if the table has data! If you _really_ want to do this, use unsafe_make_column_not_nullable")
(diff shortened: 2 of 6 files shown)


lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -133,14 +133,47 @@ def safe_change_column_default(table_name, column_name, default_value)
...
+    quoted_column_name = connection.quote_column_name(column)
+
     safely_acquire_lock_for_table(table) do
-      raw_execute "ALTER TABLE #{table} ALTER COLUMN #{column} DROP NOT NULL"
+      raw_execute "ALTER TABLE #{quoted_table_name} ALTER COLUMN #{quoted_column_name} DROP NOT NULL"
+    end
+  end
+
+  # Postgres 12+ can use a valid CHECK constraint to validate that no values of a column are null, avoiding
+  # locking the table when altering a column to NOT NULL
```
jcoleman, line 145:
```suggestion
  # Postgres 12+ can use a valid CHECK constraint to validate that no values of a column are null, avoiding
  # a full table scan while holding an exclusive lock on the table when altering a column to NOT NULL
```

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -133,14 +133,47 @@ def safe_change_column_default(table_name, column_name, default_value)
...
+  # (https://archive.is/X55up)
+  def safe_make_column_not_nullable(table, column)
+    if ActiveRecord::Base.connection.postgresql_version < 12_00_00
+      raise PgHaMigrations::InvalidMigrationError, "Cannot safely make a column non-nullable before Postgres 12"
+    end
+
+    validated_table = PgHaMigrations::Table.from_table_name(table)
+    tmp_constraint_name = "tmp_not_null_constraint_#{OpenSSL::Digest::SHA256.hexdigest(column.to_s).first(7)}"
+
+    if validated_table.has_constraint?(tmp_constraint_name)
```
jcoleman, line 158: Is it worth checking to see if there's a constraint on this column whether or not this method was the source of it? Or should we just ignore that since we weren't responsible for it?

I supposed we could add that check in when we add a method for converting an existing check constraint to a `NOT NULL` constraint...but both the `INVALID` and `VALID` cases technically are relevant here.

In fact, if we had a valid one already (not created by us even), we might just skip the check constraint creation. Not sure if that's worth it or not if we have a separate API for that.

I think I'm talking myself into "only this check is necessary for now, because it's the only thing that prevents us from proceeding, but something to chew on.

Rakefile
```diff
@@ -1,6 +1,7 @@
 require "bundler/gem_tasks"
 require "rspec/core/rake_task"
 require "appraisal"
+require "logger"
```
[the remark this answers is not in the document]
rkrage: https://github.com/braintree/pg_ha_migrations/pull/116/commits/1a97bc264d52e0732a7ebbaa29936f102d012aea

lib/pg_ha_migrations/safe_statements.rb
rkrage: Yeah I think I agree with your assessment. Would you mind opening an issue to track this new method and the enhancements described above?

jcoleman (approved):

PR #119, opened 2025-04-08
Add nulls_not_distinct support to safe_add_concurrent_partitioned_index (alternate 1)

jcoleman:
This should be supported by default in the other `add_index` methods because we already accept generic `options`, but here we need to support it specially since we define kwargs.

Note: This and the other [alternate PR](https://github.com/braintree/pg_ha_migrations/pull/120) were authored with the assistance of different LLM support. I'd like reviewers to compare the two.

### DIFF
--- a/spec/safe_statements_spec.rb
+++ b/spec/safe_statements_spec.rb
@@ -2365,6 +2365,45 @@ RSpec.describe PgHaMigrations::SafeStatements do
             test_migration.suppress_messages { test_migration.migrate(:up) }
           end
 
+          it "create an index with nulls_not_distinct option" do
+            create_range_partitioned_table(:foos3, migration_klass, with_partman: true)
+
+            test_migration = Class.new(migration_klass) do
+              def up
+                safe_add_concurrent_partitioned_index :foos3, :updated_at, nulls_not_distinct: true
+              end
+            end
+
+            allow(ActiveRecord::Base.connection).to receive(:execute).and_call_original
+
+            aggregate_failures do
+              expect(ActiveRecord::Base.connection).to receive(:execute).with(/CREATE INDEX "index_foos3_on_updated_at" ON ONLY/).once.ordered
+              expect(ActiveRecord::Base.connection).to receive(:execute).with(/CREATE INDEX CONCURRENTLY/).exactly(10).times.ordered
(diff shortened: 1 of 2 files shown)


spec/safe_statements_spec.rb
```diff
@@ -2365,6 +2365,45 @@ def up
...
+              end
+            end
+
+            allow(ActiveRecord::Base.connection).to receive(:execute).and_call_original
+
+            aggregate_failures do
+              expect(ActiveRecord::Base.connection).to receive(:execute).with(/CREATE INDEX "index_foos3_on_updated_at" ON ONLY/).once.ordered
+              expect(ActiveRecord::Base.connection).to receive(:execute).with(/CREATE INDEX CONCURRENTLY/).exactly(10).times.ordered
+              expect(ActiveRecord::Base.connection).to receive(:execute).with(/ALTER INDEX .+\nATTACH PARTITION/).exactly(10).times.ordered
+            end
```
rkrage, line 2383: This is exactly what I would have done. The expectations are way more precise and the style matches other nearby tests

_However_, I think we should also have expectations that the SQL includes `NULLS NOT DISTINCT`

spec/safe_statements_spec.rb
```diff
@@ -2365,6 +2365,45 @@ def up
...
+              tables_with_indexes.each do |table|
+                indexes = ActiveRecord::Base.connection.indexes(table)
+
+                expect(indexes.size).to eq(1)
+                expect(indexes.first).to have_attributes(
+                  table: table,
+                  name: "index_#{table}_on_updated_at",
+                  columns: ["updated_at"],
+                  using: :btree,
+                  nulls_not_distinct: true,
```
rkrage, line 2401: I guess I'm surprised that the Rails representation of an index includes this attribute. Seems like this is only available in newer versions of Rails though which is maybe why the tests are failing

All that aside, this is generally consistent with other expectations we have on index definitions

rkrage:
I actually prefer this PR description compared to the other one. This one gets straight to the point and answers the question that immediately came to my mind (why are these options only being added in _one_ of the index methods?).

jcoleman:
Closing in favor of #120.

PR #120, opened 2025-04-08
Add support for nulls_not_distinct option to index creation methods

jcoleman:
This allows users to create indexes where NULL values are treated as equal
to each other rather than always being distinct, which can be useful for
certain uniqueness constraints.

Most of the `*_add_index` methods support this by default because we
already accept generic options, but in
`safe_add_concurrent_partitioned_index`  we need to support it specially
since we define kwargs instead of taking on `options` hash argument.

In all of the methods ActiveRecord defaults to simply dropping the
option on the floor when it is not supported by the database, which we
consider to be unsafe behavior, so we update all of the `safe_*` methods
to raise helpful errors when the required invariants don't hold.

Original LLM generated message for history:

```
This change adds support for PostgreSQL 15+'s `nulls_not_distinct` index option to all methods that add indexes in the pg_ha_migrations gem. Specifically:

- Updated `safe_add_concurrent_partitioned_index` method to accept the `nulls_not_distinct` parameter and pass it through to both parent and child index creation calls

- Added a test that verifies the option is properly passed to the underlying ActiveRecord methods and that the PostgreSQL-level `indnullsnotdistinct` flag is properly set in the database

This allows users to create indexes where NULL values are treated as equal to each other rather than always being distinct, which can be useful for certain uniqueness constraints.

The implementation maintains compatibility with existing code while adding support for this PostgreSQL 15+ feature.
```

Note: This and the other [alternate PR](https://github.com/braintree/pg_ha_migrations/pull/119) were authored with the assistance of different LLM support. I'd like reviewers to compare the two.

### DIFF
--- a/lib/pg_ha_migrations/safe_statements.rb
+++ b/lib/pg_ha_migrations/safe_statements.rb
@@ -159,6 +159,16 @@ module PgHaMigrations::SafeStatements
       raise ArgumentError, "Cannot call safe_add_index_on_empty_table with :algorithm => :concurrently"
     end
 
+    # Check if nulls_not_distinct was provided but not supported
+    if options[:nulls_not_distinct] && ActiveRecord.gem_version < Gem::Version.new("7.1")
+      raise PgHaMigrations::InvalidMigrationError, "nulls_not_distinct option requires ActiveRecord 7.1 or higher"
+    end
+
+    # Check if nulls_not_distinct was provided but PostgreSQL version doesn't support it
+    if options[:nulls_not_distinct] && ActiveRecord::Base.connection.postgresql_version < 15_00_00
+      raise PgHaMigrations::InvalidMigrationError, "nulls_not_distinct option requires PostgreSQL 15 or higher"
+    end
+
     # Avoids taking out an unnecessary SHARE lock if the table does have data
     ensure_small_table!(table, empty: true)
 
@@ -171,6 +181,16 @@ module PgHaMigrations::SafeStatements
   end
 
   def safe_add_concurrent_index(table, columns, **options)
+    # Check if nulls_not_distinct was provided but not supported
+    if options[:nulls_not_distinct] && ActiveRecord.gem_version < Gem::Version.new("7.1")
+      raise PgHaMigrations::InvalidMigrationError, "nulls_not_distinct option requires ActiveRecord 7.1 or higher"
+    end
+
+    # Check if nulls_not_distinct was provided but PostgreSQL version doesn't support it
+    if options[:nulls_not_distinct] && ActiveRecord::Base.connection.postgresql_version < 15_00_00
+      raise PgHaMigrations::InvalidMigrationError, "nulls_not_distinct option requires PostgreSQL 15 or higher"
+    end
+
     unsafe_add_index(table, columns, **options.merge(:algorithm => :concurrently))
   end
 
@@ -194,8 +214,18 @@ module PgHaMigrations::SafeStatements
     using: nil,
     unique: nil,
     where: nil,
-    comment: nil
+    comment: nil,
+    nulls_not_distinct: nil
   )
+    # Check if nulls_not_distinct was provided but not supported
+    if !nulls_not_distinct.nil? && ActiveRecord.gem_version < Gem::Version.new("7.1")
+      raise PgHaMigrations::InvalidMigrationError, "nulls_not_distinct option requires ActiveRecord 7.1 or higher"
+    end
+
+    # Check if nulls_not_distinct was provided but PostgreSQL version doesn't support it
+    if !nulls_not_distinct.nil? && ActiveRecord::Base.connection.postgresql_version < 15_00_00
+      raise PgHaMigrations::InvalidMigrationError, "nulls_not_distinct option requires PostgreSQL 15 or higher"
+    end
 
     if ActiveRecord::Base.connection.postgresql_version < 11_00_00
       raise PgHaMigrations::InvalidMigrationError, "Concurrent partitioned index creation not supported on Postgres databases before version 11"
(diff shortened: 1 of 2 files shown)


spec/safe_statements_spec.rb
```diff
@@ -2518,6 +2518,49 @@ def up
...
+                )
+              end
+            end
+
+            # Check that we pass the nulls_not_distinct option to the underlying add_index method
+            expect(ActiveRecord::Base.connection).to receive(:add_index).with(
+              anything,
+              anything,
+              hash_including(nulls_not_distinct: true)
+            ).at_least(:once).and_call_original
```
rkrage, line 2540: I think we would know the exact number of times this should be called and the expectations should be more precise

spec/safe_statements_spec.rb
```diff
@@ -2518,6 +2518,49 @@ def up
...
+
+              tables_with_indexes.each do |table|
+                # Query PostgreSQL's system tables to confirm nulls_not_distinct is set
+                nulls_not_distinct = ActiveRecord::Base.connection.select_value(<<~SQL)
+                  SELECT indnullsnotdistinct
+                  FROM pg_index
+                  JOIN pg_class idx ON pg_index.indexrelid = idx.oid
+                  JOIN pg_class tbl ON pg_index.indrelid = tbl.oid
+                  WHERE tbl.relname = '#{table}'
+                  AND idx.relname LIKE 'index_%_on_text_column%'
```
rkrage, line 2556: Something like this would maybe make sense if we were backporting this feature into earlier versions of Rails. Since we're not, I prefer using the built-in Rails methods for inspecting index definitions (like we're doing in the other PR)

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -244,16 +245,23 @@ def safe_add_concurrent_partitioned_index(
...
         if_not_exists: if_not_exists,
         using: using,
         unique: unique,
         where: where,
         comment: comment,
         algorithm: :only, # see lib/pg_ha_migrations/hacks/add_index_on_only.rb
+      }
+
+      # Only include nulls_not_distinct option when supported by ActiveRecord
+      options[:nulls_not_distinct] = nulls_not_distinct if ActiveRecord.gem_version >= Gem::Version.new("7.1") && !nulls_not_distinct.nil?
```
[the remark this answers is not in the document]
jcoleman: > I'm actually curious what happens in the underlying Rails method. Will it raise an error on earlier versions of Rails when this arg is provided or does it silently ignore it?

Everything is green with the change now, so _at the very least_ having the argument passed explicitly (but not truthy) is fine. Haven't yet tested with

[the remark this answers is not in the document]
jcoleman: Earlier versions of Rails do:
```
     ArgumentError:
       Unknown key: :nulls_not_distinct. Valid keys are: :unique, :length, :order, :opclass, :where, :type, :using, :comment, :algorithm
```

spec/safe_statements_spec.rb
```diff
@@ -2518,6 +2518,51 @@ def up
...
+              expect(ActiveRecord::Base.connection).to receive(:add_index).with(
+                anything,
+                anything,
+                hash_including(nulls_not_distinct: true)
+              ).at_least(:once).and_call_original
+
+              aggregate_failures do
+                expect { test_migration.suppress_messages { test_migration.migrate(:up) } }.not_to raise_error
+
+                # Verify the nulls_not_distinct property is actually set on the created indexes
```
lmcicat, line 2546: Small nit - I don't particularly agree this comment on top of this line of code. The line is just getting the tables with indexes that will be verified below, the actual verification isn't happening on this line.

lmcicat (commented):
+1 to all of @rkrage 's feedback. One other small bit to add to it

spec/safe_statements_spec.rb
jcoleman: Yeah, I'd grant that, but that line is just setting up for the block below, so I'm inclined, if anything, to say that if personally writing it directly I'd remove the newline after this one, or even remove the temporary variable. So I'm not going to worry about it for now (there are enough other things to address with the AI changes here first).

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -221,16 +251,23 @@ def safe_add_concurrent_partitioned_index(
...
         if_not_exists: if_not_exists,
         using: using,
         unique: unique,
         where: where,
         comment: comment,
         algorithm: :only, # see lib/pg_ha_migrations/hacks/add_index_on_only.rb
+      }
+
+      # Only include nulls_not_distinct option when supported (we already checked at the top of the method)
+      options[:nulls_not_distinct] = nulls_not_distinct unless nulls_not_distinct.nil?
```
rkrage, line 265: With the check at the top of the method, I think we can just pass this through now without all this logic?

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -241,15 +278,22 @@ def safe_add_concurrent_partitioned_index(
...
+      options = {
         name: child_index.name,
         if_not_exists: if_not_exists,
         using: using,
         unique: unique,
         where: where,
+      }
+
+      # Only include nulls_not_distinct option when supported (we already checked at the top of the method)
+      options[:nulls_not_distinct] = nulls_not_distinct unless nulls_not_distinct.nil?
```
rkrage, line 290: Same comment as above

lib/pg_ha_migrations/safe_statements.rb
jcoleman: Because of removing 7.0 support, yes.

rkrage (approved):

Issue #118, opened 2025-03-06
safe_convert_check_not_null_to_not_null_constraint (or similar)


jcoleman:
There's probably a better name for the method, but we should have a method that makes it easy to convert an existing check constraint (on e.g. `column is not null` to a real `NOT NULL` constraint, since in PG12+ that's possible without a full table scan.

When we do that we should probably update `safe_make_column_not_nullable`:
- Consider checking to see if there's a constraint on this column whether or not this method was the source of it? Or should we just ignore that since we weren't responsible for it?
- If we had a valid one already (not created by us even), we might just skip the check constraint creation. Not sure if that's worth it or not if we only want the separate API for that.

See also: https://github.com/braintree/pg_ha_migrations/pull/116#discussion_r1977850714

rkrage:
Resolved by #126

PR #126, opened 2025-04-14
Add `safe_make_column_not_nullable_from_check_constraint` method

jcoleman:
This method allows safely making a column non-nullable using a validated check constraint, ensuring minimal locking and compatibility with PostgreSQL 12+. It confirms the existence and validation status of the provided constraint, raising helpful errors if either of those turn out not to be true.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,14 +189,17 @@ safe_make_column_nullable :table, :column
 ```
 #### safe\_make\_column\_not\_nullable
 
-Safely make the column not nullable - adds a temporary constraint and uses that constraint to validate no values are null before altering the column, then removes the temporary constraint
+Safely make the column not nullable - adds a temporary constraint and uses that constraint to validate no values are null before altering the column, then removes the temporary constraint.
 
 ```ruby
 safe_make_column_not_nullable :table, :column
 ```
 
-Note:
-This method runs multiple DDL statements non-transactionally. Validating the constraint could fail. In such cases an exception will be raised, and an INVALID constraint will be left on the table.
+> **Note:**
+> - This method performs a full table scan to validate that no NULL values exist in the column. While no exclusive lock is held for this scan, oon large tables the scan may take a long time.
+> - The method runs multiple DDL statements non-transactionally. Validating the constraint can fail. In such cases an exception will be raised, and an INVALID constraint will be left on the table.
+
+If you want to avoid a full table scan and have already added and validated a suitable CHECK constraint, consider using [`safe_make_column_not_nullable_from_check_constraint`](#safe_make_column_not_nullable_from_check_constraint) instead.
 
 #### unsafe\_make\_column\_not\_nullable
 
@@ -205,6 +208,23 @@ Unsafely make a column not nullable.
 ```ruby
 unsafe_make_column_not_nullable :table, :column
 ```
+
+#### safe\_make\_column\_not\_nullable\_from\_check\_constraint
+
+Variant of `safe_make_column_not_nullable` that safely makes a column NOT NULL using an existing validated CHECK constraint that enforces non-null values for the column. This method is expected to always be fast because it avoids a full table scan.
+
+```ruby
+safe_make_column_not_nullable_from_check_constraint :table, :column, constraint_name: :constraint_name
+```
+
+- `constraint_name` (required): The name of a validated CHECK constraint that enforces `column IS NOT NULL`.
+- `drop_constraint:` (optional, default: true): Whether to drop the constraint after making the column NOT NULL.
+
+You should use [`safe_make_column_not_nullable`](#safe_make_column_not_nullable_from_check_constraint) when neither a CHECK constraint or a NOT NULL constraint exists already. You should use this method when you already have an equivalent CHECK constraint on the table.
+
+This method will raise an error if the constraint does not exist, is not validated, or does not strictly enforce non-null values for the column.
+
+> **Note:**  We do not attempt to catch all possible proofs of `column IS NOT NULL` by means of an existing constraint; only a constraint with the exact definition `column IS NOT NULL` will be recognized.
 
 #### safe\_add\_index\_on\_empty\_table
 
--- a/README.md
+++ b/README.md
@@ -196,7 +196,7 @@ safe_make_column_not_nullable :table, :column
 ```
 
 > **Note:**
-> - This method performs a full table scan to validate that no NULL values exist in the column. While no exclusive lock is held for this scan, oon large tables the scan may take a long time.
+> - This method performs a full table scan to validate that no NULL values exist in the column. While no exclusive lock is held for this scan, on large tables the scan may take a long time.
 > - The method runs multiple DDL statements non-transactionally. Validating the constraint can fail. In such cases an exception will be raised, and an INVALID constraint will be left on the table.
 
 If you want to avoid a full table scan and have already added and validated a suitable CHECK constraint, consider using [`safe_make_column_not_nullable_from_check_constraint`](#safe_make_column_not_nullable_from_check_constraint) instead.
@@ -220,7 +220,7 @@ safe_make_column_not_nullable_from_check_constraint :table, :column, constraint_
 - `constraint_name` (required): The name of a validated CHECK constraint that enforces `column IS NOT NULL`.
 - `drop_constraint:` (optional, default: true): Whether to drop the constraint after making the column NOT NULL.
 
-You should use [`safe_make_column_not_nullable`](#safe_make_column_not_nullable_from_check_constraint) when neither a CHECK constraint or a NOT NULL constraint exists already. You should use this method when you already have an equivalent CHECK constraint on the table.
+You should use [`safe_make_column_not_nullable`](#safe_make_column_not_nullable) when neither a CHECK constraint or a NOT NULL constraint exists already. You should use this method when you already have an equivalent CHECK constraint on the table.
 
 This method will raise an error if the constraint does not exist, is not validated, or does not strictly enforce non-null values for the column.
 
--- a/lib/pg_ha_migrations/constraint.rb
+++ b/lib/pg_ha_migrations/constraint.rb
@@ -1,0 +1,1 @@
+PgHaMigrations::CheckConstraint = Struct.new(:name, :definition, :validated)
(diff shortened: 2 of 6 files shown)


jcoleman:
Fixes #118.

From the original issue:

When we do that we should probably update `safe_make_column_not_nullable`:

- Consider checking to see if there's a constraint on this column whether or not this method was the source of it? Or should we just ignore that since we weren't responsible for it?
- If we had a valid one already (not created by us even), we might just skip the check constraint creation. Not sure if that's worth it or not if we only want the separate API for that.

I've left that as well as whether we should update the existing `safe_make_column_not_nullable` to use this new method internally for discussion on this PR.

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -168,6 +168,42 @@ def safe_make_column_not_nullable(table, column)
...
+    end
+
+    unless constraint_name
+      raise ArgumentError, "Expected <constraint_name> to be present"
+    end
+
+    quoted_table_name = connection.quote_table_name(table)
+    quoted_column_name = connection.quote_column_name(column)
+
+    constraint = ActiveRecord::Base.connection.select_one(<<~SQL)
```
jcoleman, line 183: I just noticed the existence of `PgHaMigrations::Table.has_constraint?`; that might imply we should standardize for use here as well.

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -168,6 +168,42 @@ def safe_make_column_not_nullable(table, column)
...
+
+    unless constraint["convalidated"]
+      raise PgHaMigrations::InvalidMigrationError, "The provided constraint is not validated"
+    end
+
+    unless constraint["constraint_def"] =~ /\ACHECK \(*(#{Regexp.escape(column.to_s)}|#{Regexp.escape(quoted_column_name)}) IS NOT NULL\)*\Z/i
+      raise PgHaMigrations::InvalidMigrationError, "The provided constraint does not enforce non-null values for the column"
+    end
+
+    safely_acquire_lock_for_table(table) do
```
jcoleman, line 202: `unsafe_make_column_not_nullable` already does this locking.

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -168,6 +168,42 @@ def safe_make_column_not_nullable(table, column)
...
+      unsafe_make_column_not_nullable(table, column)
```
imsurajmishra, line 203: Just want to callout that the reason this works is because 

From postgres [docs](https://www.postgresql.org/docs/current/sql-altertable.html):

> SET NOT NULL may only be applied to a column provided none of the records in the table contain a NULL value for the column. Ordinarily this is checked during the ALTER TABLE by scanning the entire table; however, if a valid CHECK constraint is found which proves no NULL can exist, then the table scan is skipped.

lib/pg_ha_migrations/safe_statements.rb
imsurajmishra: That currently just checks if the constraint exists, but here we need more stuff for additional verification like convalidated, constraint_def.
Probably need to modify that method if we need to return these values.

lib/pg_ha_migrations/safe_statements.rb
jcoleman: I added this to the diff.

lib/pg_ha_migrations/safe_statements.rb
jcoleman: See the updated diff.

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -150,10 +150,51 @@ def safe_make_column_not_nullable(table, column)
...
+
+    unless constraint
+      raise PgHaMigrations::InvalidMigrationError, "The provided constraint does not exist"
+    end
+
+    unless constraint.validated
+      raise PgHaMigrations::InvalidMigrationError, "The provided constraint is not validated"
+    end
+
+    unless constraint.definition =~ /\ACHECK \(*(#{Regexp.escape(column.to_s)}|#{Regexp.escape(quoted_column_name)}) IS NOT NULL\)*\Z/i
```
jcoleman, line 187: We might want to add a test to ensure this doesn't catch constraints that, e.g., have the form `a IS NOT NULL OR b IS NOT NULL`.

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -150,10 +150,58 @@ def safe_make_column_not_nullable(table, column)
...
+    # `a IS NOT NULL`.
+    unless constraint.definition =~ /\ACHECK \(*(#{Regexp.escape(column.to_s)}|#{Regexp.escape(quoted_column_name)}) IS NOT NULL\)*\Z/i
+      raise PgHaMigrations::InvalidMigrationError, "The provided constraint does not enforce non-null values for the column"
+    end
+
+    # "Ordinarily this is checked during the ALTER TABLE by scanning the entire table; however, if a
+    # valid CHECK constraint is found which proves no NULL can exist, then the table scan is
+    # skipped."
+    # See: https://www.postgresql.org/docs/current/sql-altertable.html#SQL-ALTERTABLE-DESC-SET-DROP-NOT-NULL
+    unsafe_make_column_not_nullable(table, column)
```
rkrage, line 202: Should we provide an optional arg to additionally drop the check constraint?

lib/pg_ha_migrations/constraint.rb
```diff
@@ -0,0 +1 @@
+PgHaMigrations::CheckConstraint = Struct.new(:name, :definition, :validated)
```
rkrage, line 1: I understand why this isn't in the `Relation` file, but is there somewhere else we can put this? A single line file seems strange to my brain, but obviously this isn't a _problem_

jcoleman: We could put it right above the constraints method, but...it's just a question of whether we expect to move it later IMO.

README.md
```diff
@@ -189,14 +189,17 @@ safe_make_column_nullable :table, :column
...
+Safely make the column not nullable - adds a temporary constraint and uses that constraint to validate no values are null before altering the column, then removes the temporary constraint.
 
 ```ruby
 safe_make_column_not_nullable :table, :column
 ```
 
-Note:
-This method runs multiple DDL statements non-transactionally. Validating the constraint could fail. In such cases an exception will be raised, and an INVALID constraint will be left on the table.
+> **Note:**
+> - This method performs a full table scan to validate that no NULL values exist in the column. While no exclusive lock is held for this scan, oon large tables the scan may take a long time.
```
rkrage, line 199:
```suggestion
> - This method performs a full table scan to validate that no NULL values exist in the column. While no exclusive lock is held for this scan, on large tables the scan may take a long time.
```

README.md
```diff
@@ -206,6 +209,23 @@ Unsafely make a column not nullable.
...
+Variant of `safe_make_column_not_nullable` that safely makes a column NOT NULL using an existing validated CHECK constraint that enforces non-null values for the column. This method is expected to always be fast because it avoids a full table scan.
+
+```ruby
+safe_make_column_not_nullable_from_check_constraint :table, :column, constraint_name: :constraint_name
+```
+
+- `constraint_name` (required): The name of a validated CHECK constraint that enforces `column IS NOT NULL`.
+- `drop_constraint:` (optional, default: true): Whether to drop the constraint after making the column NOT NULL.
+
+You should use [`safe_make_column_not_nullable`](#safe_make_column_not_nullable_from_check_constraint) when neither a CHECK constraint or a NOT NULL constraint exists already. You should use this method when you already have an equivalent CHECK constraint on the table.
```
rkrage, line 223:
```suggestion
You should use [`safe_make_column_not_nullable`](#safe_make_column_not_nullable) when neither a CHECK constraint or a NOT NULL constraint exists already. You should use this method when you already have an equivalent CHECK constraint on the table.
```

rkrage (approved):
A few lingering comments / suggestions but this is really close

lib/pg_ha_migrations/safe_statements.rb
jcoleman: Done.

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -150,10 +150,69 @@ def safe_make_column_not_nullable(table, column)
...
+    safely_acquire_lock_for_table(table) do
+      # "Ordinarily this is checked during the ALTER TABLE by scanning the entire table; however, if a
+      # valid CHECK constraint is found which proves no NULL can exist, then the table scan is
+      # skipped."
+      # See: https://www.postgresql.org/docs/current/sql-altertable.html#SQL-ALTERTABLE-DESC-SET-DROP-NOT-NULL
+      unsafe_make_column_not_nullable(table, column)
+
+      if drop_constraint
+        unsafe_remove_constraint(table, name: constraint_name)
+      end
```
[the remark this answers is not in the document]
jcoleman: Heh, could be, yes, though I'm not particularly a fan of that style (in this case it'd also be 81 characters, beyond our standard 80 character line length :D

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -141,7 +141,7 @@ def safe_make_column_not_nullable(table, column)
     validated_table = PgHaMigrations::Table.from_table_name(table)
     tmp_constraint_name = "tmp_not_null_constraint_#{OpenSSL::Digest::SHA256.hexdigest(column.to_s).first(7)}"
 
-    if validated_table.has_constraint?(tmp_constraint_name)
+    if validated_table.check_constraints.any? { |c| c.name == tmp_constraint_name }
       raise PgHaMigrations::InvalidMigrationError, "A constraint #{tmp_constraint_name.inspect} already exists. " \
         "This implies that a previous invocation of this method failed and left behind a temporary constraint. " \
         "Please drop the constraint before attempting to run this method again."
```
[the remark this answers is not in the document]
jcoleman: I decided to make the following two issues instead for follow-up; I think they represent a better wya to handle this than "raise and error and make the user deal with it".

https://github.com/braintree/pg_ha_migrations/issues/130
https://github.com/braintree/pg_ha_migrations/issues/131

PR #133, opened 2025-04-18
Split safe statements test file for speed

jcoleman:
The lock acquisition tests are particularly slow, and we don't modify
those methods often, so it simplifies local development to split these
out.


spec/utility_methods_spec.rb
```diff
@@ -0,0 +1,245 @@
+require "spec_helper"
+
+RSpec.describe PgHaMigrations::SafeStatements, "utility methods" do
```
rkrage, line 3: Does `safe_set_maintenance_work_mem_gb` fall into this category as well?

spec/spec_helper.rb
```diff
@@ -76,4 +76,98 @@ def self.drop_objects
       ActiveRecord::Base.connection.execute("DROP TYPE #{enum} CASCADE")
     end
   end
+
+  TableLock = Struct.new(:table, :lock_type, :granted, :pid)
+  def self.locks_for_table(table, connection:)
```
rkrage, line 81: Is it worth putting `DatabaseHelper` in a different file at this point?

rkrage:
A few general questions:
- Do we want separate spec files for safe / unsafe statements?
- Do we want to actually split out the definitions for the utility methods into a separate file?

spec/utility_methods_spec.rb
jcoleman: Yes, it does, I'll move that also.

spec/spec_helper.rb
jcoleman: I was wondering that; I was also wondering if it should be split between "used by tests" and "used by spec_helper.rb".

Thoughts?

rkrage: I like the idea of splitting them up. I don't think the "used by spec helper" methods necessarily need to be in a different file though. Might make sense to inline those, but I don't feel too strongly

spec/utility_methods_spec.rb
jcoleman: Done.

spec/spec_helper.rb
jcoleman: Put the new stuff in a new file.

rkrage (approved):

PR #134, opened 2025-04-22
Add idempotence and existing constraint re-use to safe_make_column_not_nullable

### LINKED ISSUE #130: safe_make_column_not_nullable should idempotently handle a failed constraint validation
This method creates a temporary CHECK constraint, and validation of that constraint can fail non-transactionally. If it does we should either clean up after ourselves, and/or (probably need to do this even if we also do the cleanup so we handle crashes of the migration runner as well) be able to retry validating an existing constraint.

### LINKED ISSUE #131: safe_make_column_not_nullable should be able to rely on an existing CHECK constraint if it exists
This would be a convenience, given we have `safe_make_column_not_nullable_from_check_constraint` once https://github.com/braintree/pg_ha_migrations/pull/126 is merged, but it seems worth doing as it's simple and would maintain coherency across the methods.

This is distinct from, but certainly overlapping with https://github.com/braintree/pg_ha_migrations/issues/130

jcoleman:
If an existing constraint already exists we should re-use that. If it's not yet validated, we still need to validate it -- that gives use idempotency, and we no longer need to raise an error if the previous run failed. If the existing constraint is already validated, there's no need to validate it, and we can simply proceed.

While it's true that _if you already know you have a validated constraint_ we idiomatically prefer using
`safe_make_column_not_nullable_from_check_constraint` in cases where your database state might be unknown (e.g., a previous failure), it's still helpful to be able to use `safe_make_column_not_nullable`.

Fixes #130
Fixes #131

### DIFF
--- a/lib/pg_ha_migrations/safe_statements.rb
+++ b/lib/pg_ha_migrations/safe_statements.rb
@@ -139,23 +139,39 @@ module PgHaMigrations::SafeStatements
     end
 
     validated_table = PgHaMigrations::Table.from_table_name(table)
-    tmp_constraint_name = "tmp_not_null_constraint_#{OpenSSL::Digest::SHA256.hexdigest(column.to_s).first(7)}"
+    quoted_column_name = connection.quote_column_name(column)
+    column_str = column.to_s
 
-    if validated_table.check_constraints.any? { |c| c.name == tmp_constraint_name }
-      raise PgHaMigrations::InvalidMigrationError, "A constraint #{tmp_constraint_name.inspect} already exists. " \
-        "This implies that a previous invocation of this method failed and left behind a temporary constraint. " \
-        "Please drop the constraint before attempting to run this method again."
+    # First, look for existing constraints that match the IS NOT NULL pattern for this column
+    existing_constraint = validated_table.check_constraints.select do |c|
+      c.definition =~ /\ACHECK \(*(#{Regexp.escape(column_str)}|#{Regexp.escape(quoted_column_name)}) IS NOT NULL\)*\Z/i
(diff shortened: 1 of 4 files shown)


rkrage (approved):

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -139,23 +139,39 @@ def safe_make_column_not_nullable(table, column)
...
+    column_str = column.to_s
 
-    if validated_table.check_constraints.any? { |c| c.name == tmp_constraint_name }
-      raise PgHaMigrations::InvalidMigrationError, "A constraint #{tmp_constraint_name.inspect} already exists. " \
-        "This implies that a previous invocation of this method failed and left behind a temporary constraint. " \
-        "Please drop the constraint before attempting to run this method again."
-    end
+    # First, look for existing constraints that match the IS NOT NULL pattern for this column
+    existing_constraint = validated_table.check_constraints.select do |c|
+      c.definition =~ /\ACHECK \(*(#{Regexp.escape(column_str)}|#{Regexp.escape(quoted_column_name)}) IS NOT NULL\)*\Z/i
```
rkrage, line 147: I did just approve this but it just occurred to me that maybe we want specific tests for this regex? Like, it looks right to me, but should we write a test that makes sure this method doesn't discover a constraint that has a `IS NOT NULL AND ...` kind of definition?

jcoleman: Yeah, I think we have those tests for `safe_make_column_not_nullable_from_check_constraint` already; I'll add those before merging.

PR #147, opened 2026-07-21
Make safe_add_concurrent_partitioned_index resumable after a partial failure

alexspeller:
## Problem

`safe_add_concurrent_partitioned_index(..., if_not_exists: true)` is documented/intended to be resumable, but it is **not** resumable after a run that fails partway through building the per-partition indexes.

When a `CREATE INDEX CONCURRENTLY` on one partition dies (statement timeout, lock timeout, connection drop, deploy interruption…), it leaves an **unattached, `indisvalid = false`** index behind on that child partition. On the next run:

1. The parent `ON ONLY` create (`IF NOT EXISTS`) is skipped by name — fine.
2. For the child, `CREATE INDEX CONCURRENTLY IF NOT EXISTS <name>` matches on **name only** — Postgres never checks index validity — so the invalid leftover is skipped and **never rebuilt**.
3. The attach step attaches the still-invalid child, so the parent `ON ONLY` index never becomes valid.
4. The method hits its final assertion and raises `"Unexpected state. Parent index ... is invalid"`.

Because nothing rebuilds the invalid child, this raises on **every** subsequent run — the migration never converges without manual `DROP INDEX` cleanup.

The existing "creates valid index when partially created" spec only exercises the case where children are *missing* (which `IF NOT EXISTS` rebuilds fine); an *invalid* leftover child is a different, uncovered case.

## Fix

When resuming with `if_not_exists`, drop any invalid child index leftovers before the (re)build so they get rebuilt and the parent can validate. The invalid children are guaranteed **unattached** at this point — the attach step runs only after every child index has been built — so they can be dropped concurrently via the existing `safe_remove_concurrent_index`. The detection matches the exact child index names for the current partitions (and `relkind = 'i'`, so sub-partition parent indexes are left to the recursive call), and is a single catalog query gated on `if_not_exists`, so the non-resume paths are unchanged.

## Test

Adds a spec that simulates the real failure state (one child valid + attached, one child built-but-invalid + unattached, parent invalid) and asserts that a re-run with `if_not_exists: true` drops the invalid leftover, rebuilds it, and converges to a fully valid parent index. Without the fix this spec fails with the exact `"Parent index ... is invalid"` raise.

## Verification

Ran the affected `#safe_add_concurrent_partitioned_index` suite locally against the project's `docker compose` Postgres 18 / pg_partman 5 image (`bundle exec rspec spec/safe_statements_spec.rb -e safe_add_concurrent_partitioned_index`), green across all `ActiveRecord::Migration::Compatibility` versions. Confirmed the new spec is red without the change and green with it. CI matrix will exercise the full PG/partman/Rails grid.

### DIFF
--- a/lib/pg_ha_migrations/safe_statements.rb
+++ b/lib/pg_ha_migrations/safe_statements.rb
@@ -346,7 +346,7 @@ module PgHaMigrations::SafeStatements
           AND pg_class.relname = ANY (ARRAY[#{quoted_child_index_names}])
       SQL
 
-      invalid_child_index_names.each do |child_index|
+      child_indexes.select { |n| invalid_child_index_names.include?(n) }.each do |child_index|
         safe_remove_concurrent_index(child_index.table.fully_qualified_name, name: child_index.name)
       end
     end
--- a/lib/pg_ha_migrations/safe_statements.rb
+++ b/lib/pg_ha_migrations/safe_statements.rb
@@ -346,9 +346,7 @@ module PgHaMigrations::SafeStatements
           AND pg_class.relname = ANY (ARRAY[#{quoted_child_index_names}])
       SQL
 
-      child_indexes.each do |child_index|
-        next unless invalid_child_index_names.include?(child_index.name)
-
+      invalid_child_index_names.each do |child_index|
         safe_remove_concurrent_index(child_index.table.fully_qualified_name, name: child_index.name)
       end
     end
--- a/lib/pg_ha_migrations/safe_statements.rb
+++ b/lib/pg_ha_migrations/safe_statements.rb
@@ -326,6 +326,33 @@ module PgHaMigrations::SafeStatements
       PgHaMigrations::Index.from_table_and_columns(child_table, columns)
     end
 
+    # A previous run that was interrupted mid-build can leave an invalid index
+    # behind on a child partition. Because CREATE INDEX ... IF NOT EXISTS matches
+    # on name only (not validity), that leftover would be skipped and never
+    # rebuilt, so the parent index would never become valid and this method
+    # would raise below on every re-run. When resuming (if_not_exists), drop any
+    # invalid child leftovers first so they get rebuilt. They are unattached at
+    # this point (the attach step runs only after every child index is built),
+    # so they can be dropped concurrently.
+    if if_not_exists && child_indexes.present?
+      quoted_child_index_names = child_indexes.map { |child_index| connection.quote(child_index.name) }.join(", ")
+
+      invalid_child_index_names = connection.select_values(<<~SQL)
+        SELECT pg_class.relname
+        FROM pg_index
+          JOIN pg_class ON pg_class.oid = pg_index.indexrelid
+        WHERE NOT pg_index.indisvalid
+          AND pg_class.relkind = 'i'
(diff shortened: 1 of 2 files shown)


jcoleman:
@alexspeller I see this is in draft; do you want us to review it now? Or is there still work being done?

alexspeller:
oops I left it in draft whilst I was checking it worked through a prod deploy for us - it did work fine so hopefully this is ready to merge now!

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -326,6 +326,33 @@ def safe_add_concurrent_partitioned_index(
...
+        FROM pg_index
+          JOIN pg_class ON pg_class.oid = pg_index.indexrelid
+        WHERE NOT pg_index.indisvalid
+          AND pg_class.relkind = 'i'
+          AND pg_class.relname = ANY (ARRAY[#{quoted_child_index_names}])
+      SQL
+
+      child_indexes.each do |child_index|
+        next unless invalid_child_index_names.include?(child_index.name)
+
```
jcoleman:
```suggestion
      invalid_child_index_names.each do |child_index|
```

jcoleman (commented):
One small suggestion; I think it's equivalent just simpler.

lib/pg_ha_migrations/safe_statements.rb
jcoleman: Oops, this is a string vs. the index object; I didn't catch that. Will fix in a minute.

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -326,6 +326,31 @@ def safe_add_concurrent_partitioned_index(
...
+      invalid_child_index_names = connection.select_values(<<~SQL)
+        SELECT pg_class.relname
+        FROM pg_index
+          JOIN pg_class ON pg_class.oid = pg_index.indexrelid
+        WHERE NOT pg_index.indisvalid
+          AND pg_class.relkind = 'i'
+          AND pg_class.relname = ANY (ARRAY[#{quoted_child_index_names}])
+      SQL
+
+      invalid_child_index_names.each do |child_index|
```
jcoleman:
```suggestion
      child_indexes.select { |n| invalid_child_index_names.include?(n) }.each do |child_index|
```

lib/pg_ha_migrations/safe_statements.rb
```diff
@@ -326,6 +326,31 @@ def safe_add_concurrent_partitioned_index(
...
+      invalid_child_index_names = connection.select_values(<<~SQL)
+        SELECT pg_class.relname
+        FROM pg_index
+          JOIN pg_class ON pg_class.oid = pg_index.indexrelid
+        WHERE NOT pg_index.indisvalid
+          AND pg_class.relkind = 'i'
+          AND pg_class.relname = ANY (ARRAY[#{quoted_child_index_names}])
+      SQL
+
+      child_indexes.select { |n| invalid_child_index_names.include?(n) }.each do |child_index|
```
jcoleman:
```suggestion
      child_indexes.select { |index| invalid_child_index_names.include?(index.name) }.each do |child_index|
```

jcoleman (approved):

jcoleman:
Pushed gem version 2.3.1 with this fix.

